Issue #83, opened 2014-05-30
Insatller: Data Fixtures Fail


astorm:
After installing the OroBAP platform via composer and loading the web installer in a browser (on May 30th, 2014, making this presumably the 1.2 release), the installer fails to load its fixtures at step 3

![screen shot 2014-05-30 at 12 17 41 pm](https://cloud.githubusercontent.com/assets/184372/3135162/1c60ba94-e82f-11e3-8a37-f1cdeb9e4f1e.png)

The following error appears in the apache logs at the exact moment the data fixtures step fails

```
[Fri May 30 23:13:33 2014] [error] [client 71.237.181.69] PHP Warning:  Erroneous data format for unserializing 'Oro\\Bundle\\UserBundle\\Entity\\User' in /home/astorm/public_html/testbed.pulsestorm.net/public/platform-application/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php on line 869, referer: http://testbed.pulsestorm.net/installer/flow/oro_installer/schema
[Fri May 30 23:13:33 2014] [error] [client 71.237.181.69] PHP Notice:  unserialize(): Error at offset 44 of 45 bytes in /home/astorm/public_html/testbed.pulsestorm.net/public/platform-application/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php on line 869, referer: http://testbed.pulsestorm.net/installer/flow/oro_installer/schema
[Fri May 30 23:13:33 2014] [error] [client 71.237.181.69] PHP Fatal error:  __clone method called on non-object in /home/astorm/public_html/testbed.pulsestorm.net/public/platform-application/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php on line 872, referer: http://testbed.pulsestorm.net/installer/flow/oro_installer/schema
```

PR #184, opened 2014-11-18
Update User.php

kmuenze:
updates to support adding Phone numbers to users. issue #83 in orocrm/crm.


DimaSoroka:
Closing this PR as phone was added to the User entity as extended field.

PR #185, opened 2014-11-18
Update UserManager.php

kmuenze:
update to enable search for users by phone number. (https://github.com/orocrm/crm/issues/84)


zebimax:
Class BaseUserManager which now extends UserManager class has findUserBy method which can be used to find one user by special criteria(including 'phone'). Also phone is not unique field for user now so search one user by phone is not correct operation.

PR #217, opened 2015-01-15
Google SSO

nenadalm:
Once this PR is merged, I'll send another to crm-application and platform-application with updated parameters.yml.dist:
google.client_id: ~
google.client_secret: ~

I think it would be also nice to include phpunit into dev-dependencies (so people don't have to install it globally and struggle with versions)

I suppose it's not necessary to write doc for this PR...


src/Oro/Bundle/SecurityBundle/Authentication/Token/OrganizationContextTokenTrait.php
```diff
@@ -0,0 +1,27 @@
...
+
+trait OrganizationContextTokenTrait
+{
+    /** @var  Organization */
+    protected $organization;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getOrganizationContext()
```
yurio: Please add comments what this function do. {@inheritdoc} here is not good here because you do not extend any class, etc. For example, you can see https://github.com/orocrm/platform/blob/master/src/Oro/Bundle/ActivityBundle/Model/ExtendActivity.php

src/Oro/Bundle/SecurityBundle/Authentication/Token/OrganizationContextTokenTrait.php
```diff
@@ -0,0 +1,27 @@
...
+    {
+        return $this->organization;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setOrganizationContext(Organization $organization)
```
yurio: Please add comments what this function do. {@inheritdoc} here is not good here because you do not extend any class, etc. For example, you can see https://github.com/orocrm/platform/blob/master/src/Oro/Bundle/ActivityBundle/Model/ExtendActivity.php

src/Oro/Bundle/UserBundle/Entity/User.php
```diff
@@ -1723,4 +1730,20 @@ public function getOrganization()
...
+     */
+    public function getGoogleId()
+    {
+        return $this->googleId;
+    }
+
+    /**
+     * @param string|null $googleId
+     */
+    public function setGoogleId($googleId = null)
```
yurio: All the setters in this class has `return $this;`. Why this method has no it?
Please add short description what this medod do.

src/Oro/Bundle/UserBundle/Entity/User.php
```diff
@@ -1723,4 +1730,20 @@ public function getOrganization()
     {
         return $this->organization;
     }
+
+    /**
+     * @return string|null
+     */
+    public function getGoogleId()
```
yurio: Please add short description what this medod do.

src/Oro/Bundle/UserBundle/Entity/UserManager.php
```diff
@@ -32,17 +38,23 @@ class UserManager implements UserProviderInterface
...
+     * @var ConfigManager
+     */
+    protected $cm;
+
+    /**
      * Constructor
      *
      * @param string                  $class          Entity name
      * @param ObjectManager           $om             Object manager
      * @param EncoderFactoryInterface $encoderFactory
```
yurio: Please add $cm to parameters list

src/Oro/Bundle/UserBundle/Entity/UserManager.php
```diff
@@ -314,4 +326,71 @@ public function getApi(User $user, Organization $organization)
...
+        if (!$this->cm->get('oro_user.enable_google_sso')) {
+            throw new \Exception('SSO is not enabled');
+        }
+
+        $username = $response->getUsername();
+        if ($username === null) {
+            throw new AccountNotLinkedException(sprintf("User '%s' not found.", $username));
+        }
+
+        $user = $this->findUserBy(array($this->getOAuthProperty($response) => $username));
```
yurio: you can use short array syntax here

src/Oro/Bundle/UserBundle/Entity/UserManager.php
```diff
@@ -314,4 +326,71 @@ public function getApi(User $user, Organization $organization)
...
+     * @return boolean
+     */
+    protected function isEmailEnabledForOauth($email)
+    {
+        $enabledDomain = $this->cm->get('oro_user.google_sso_domain');
+        if (!$enabledDomain) {
+            return true;
+        }
+
+        return (bool) preg_match(sprintf('/@%s$/', $enabledDomain), $email);
```
yurio: As i see, parameter  'oro_user.google_sso_domain' can not take multiple domains. I think it's better to allow to configure multiple domains. What do you think?

src/Oro/Bundle/UserBundle/Migrations/Schema/v1_9/OroUserBundle.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace Oro\Bundle\UserBundle\Migrations\Schema\v1_9;
```
yurio: Please change migration version because in nearest release we have migration v 1_9

yurio:
Please update Oro\Bundle\UserBundle\Migrations\Schema\OroUserBundleInstaller with your changes

src/Oro/Bundle/UserBundle/Security/OAuthProvider.php
```diff
@@ -0,0 +1,71 @@
...
+     * @param UserCheckerInterface $userChecker User checker
+     */
+    public function __construct(OAuthAwareUserProviderInterface $userProvider, ResourceOwnerMap $resourceOwnerMap, UserCheckerInterface $userChecker)
+    {
+        $this->userProvider = $userProvider;
+        $this->resourceOwnerMap = $resourceOwnerMap;
+        $this->userChecker = $userChecker;
+    }
+
+    public function authenticate(TokenInterface $token)
```
yurio: Please add short description what this method do

src/Oro/Bundle/UserBundle/Security/OAuthProvider.php
```diff
@@ -0,0 +1,71 @@
...
+use HWI\Bundle\OAuthBundle\Security\Core\Authentication\Provider\OAuthProvider as HWIOAuthProvider;
+use HWI\Bundle\OAuthBundle\Security\Core\Exception\OAuthAwareExceptionInterface;
+use HWI\Bundle\OAuthBundle\Security\Core\User\OAuthAwareUserProviderInterface;
+use HWI\Bundle\OAuthBundle\Security\Http\ResourceOwnerMap;
+
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Core\User\UserCheckerInterface;
+
+/**
+ * @author Miloslav Nenadal <miloslav.nenadal@imatic.cz>
```
yurio: We does not use this annotation

src/Oro/Bundle/UserBundle/Security/OAuthProvider.php
```diff
@@ -0,0 +1,71 @@
...
+        $this->resourceOwnerMap = $resourceOwnerMap;
+        $this->userChecker = $userChecker;
+    }
+
+    public function authenticate(TokenInterface $token)
+    {
+        /* @var OAuthToken $token */
+        $resourceOwner = $this->resourceOwnerMap->getResourceOwnerByName($token->getResourceOwnerName());
+
+        $userResponse = $resourceOwner->getUserInformation($token->getRawToken());
```
yurio: Do you need this variable here? You use it only once at line 53

src/Oro/Bundle/UserBundle/Entity/User.php
```diff
@@ -359,6 +359,13 @@ class User extends ExtendUser implements
     protected $loginCount;
 
     /**
+     * @var string
+     *
+     * @ORM\Column(name="google_id", type="string", nullable=true)
+     */
+    protected $googleId;
```
yurio: Please add unit tests for getters and setters

yurio:
Please add tests for OAuthProvider

yurio:
Please add short documentation about ligin ability via Google OAuth

src/Oro/Bundle/UserBundle/Security/OAuthProvider.php
```diff
@@ -0,0 +1,71 @@
...
+        } catch (OAuthAwareExceptionInterface $e) {
+            $e->setToken($token);
+            $e->setResourceOwnerName($token->getResourceOwnerName());
+
+            throw $e;
+        }
+
+        $token = new OAuthToken($token->getRawToken(), $user->getRoles());
+        $token->setResourceOwnerName($resourceOwner->getName());
+        $token->setOrganizationContext($user->getOrganization());
```
yurio: It's not good to use $user->getOrganization() here because thereis case that user has no permittion to organization where this user was created.

src/Oro/Bundle/UserBundle/Entity/UserManager.php
nenadalm: I saw that you're using short array syntax just sometimes (probably newer code?) so I wrote it according to the class.

So you're using short syntax for all new code in all places?

src/Oro/Bundle/UserBundle/Entity/UserManager.php
nenadalm: Sure - I think it would be better (even if it would be enough in most cases to have one domain probably).

So I'll add support for multiple domains.

src/Oro/Bundle/UserBundle/Migrations/Schema/v1_9/OroUserBundle.php
nenadalm: Ok. So I'll change it to 1_10.

src/Oro/Bundle/UserBundle/Security/OAuthProvider.php
nenadalm: sorry - template from IDE put it here - I missed it somehow.

src/Oro/Bundle/UserBundle/Entity/UserManager.php
yurio: yes, for new code we use short array syntax

src/Oro/Bundle/UserBundle/Entity/UserManager.php
nenadalm: fixed https://github.com/orocrm/platform/pull/217/files#diff-59801e8c0741b3fe8f27b2036c996838R51

src/Oro/Bundle/UserBundle/Entity/User.php
nenadalm: fixed https://github.com/orocrm/platform/pull/217/files#diff-c0b3045f465ab49e0465bdc15bb125a7R395

nenadalm:
everything should be ok now if build passes.

src/Oro/Bundle/ConfigBundle/Resources/config/system_configuration.yml
```diff
@@ -7,6 +7,9 @@ oro_system_configuration:
         general_setup:
             title: oro.config.system_configuration.groups.general_setup.title
 
+        integration_setup:
```
yurio: Why did you add this group here? I think it should be in  src/Oro/Bundle/UserBundle/Resources/config/system_configuration.yml

src/Oro/Bundle/ConfigBundle/Resources/translations/messages.en.yml
```diff
@@ -20,6 +20,7 @@ oro:
             groups:
                 platform.title: "System configuration"
                 general_setup.title: "General setup"
+                integration_setup.title: "Integration"
```
yurio: Why did you add this key here? Please move it to user bundle

src/Oro/Bundle/ConfigBundle/Resources/config/system_configuration.yml
nenadalm: Because I noticed you have basic group (general_setup) inside ConfigBundle.
integration_setup seems to be another general group which could use more bundles than UserBundle.

So I thought ConfigBundle is the place. But I'll move it. (the same for massages below).

nenadalm:
all fixed

yurio:
Thanks for your fixes.
But, then i try to run platform with your changes, i got error 'Fatal error: Uncaught exception 'Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException' with message 'The service "hwi_oauth.security.oauth_utils" has a dependency on a non-existent service "hwi_oauth.resource_ownermap.main"'.

 google.client_id and google.client_secret parameters was set in parameters.yml. What i missed else?

And one thing. What do you think about to put this changes to the separate bundle? In this case, we can do extend package for Oro from this bundle.

nenadalm:
it works for me (only difference is security configuration probably):
https://github.com/nenadalm/platform-application/compare/google_sso?expand=1

related to the separate bundle:
* google sso sounds to me like something which doesn't have to be necessarily inside basic setup and could be installed as a plugin
* if you want to have google sso in basic setup, it could stay here I think - as it is related with users and the functionality is inside UserBundle (sure you could think of it as separate feature, but in case you don't need/want it as separate feature, I don't think it's necessary)

nenadalm:
if you want it as separate bundle (even repository?) tell me and I'll fix it.

sergeyz:
@nenadalm Hi! You can add an extension i think /cc @DimaSoroka

iklymenko:
From my point of view, would be great to have this functionality like separate extension.

DimaSoroka:
@nenadalm, please move it separate bundle. Would be great to have this as a separate package, but this would require to move security.yml update to bundle level.

nenadalm:
sure - will do tomorrow. So by separate bundle you mean the same repository as I understand.

DimaSoroka:
At list separate folder so we can make it optional later on. The idea is that we should be able to make this bundle optional in the distribution and bundle should not require files update on application level.

nenadalm:
Now it should be ok.

Regarding to application level files:
```Fatal error: Uncaught exception 'Symfony\Component\Config\Definition\Exception\InvalidConfigurationException' with message 'You are not allowed to define new elements for path "security.firewalls". Please define all elements for this path in one config file.```
it doesn't seems to be possible to configure security.yml by prepending configuration in standard way. So I could investigate this in case you don't have such feature already.

DimaSoroka:
Please check what we can do for security.yml, this will be very helpful in future.

nenadalm:
It could be solved using: https://github.com/orocrm/platform/commit/23cc6d6741a50eb8a3b5ab584df3bd35ff569b3d

in case you agree with it - I'll add a few tests.

nenadalm:
Since there are no remarks, I assume you like it so I sent the tests and the PR is ready to merge. (in case tests will pass ofc)

yurio:
Hi, @nenadalm.
Thanks for the changes. But I was expected that security firewall config will be done as separate yml file. In this case it will be the same as app.yml files from our bundles. With this files we can change main config.yml.
Example of usage this files can be found at https://github.com/orocrm/platform/blob/master/src/Oro/Bundle/AttachmentBundle/Resources/config/oro/app.yml . In this example we configure knp_gaufrette bundle from another (attachment) bundle.

nenadalm:
done https://github.com/nenadalm/platform/blob/7d6c68f5ab9c44cca5ae4ecb9cb68a79e91a1c1d/src/Oro/Bundle/SSOBundle/Resources/config/oro/app.yml

iklymenko:
Hi, few more comments:
1. SSO must be optional at the moment google.client_id and google.secret are mandatory in parameters.yml and can not be ~
2. Application installation is failing, because SSO bundle requires user tables but this bundle was included before UserBundle.
3. Would be great to have a link in documentation to Google API doc - how to configure Google OAuth.

nenadalm:
> 1. SSO must be optional at the moment google.client_id and google.secret are mandatory in parameters.yml and can not be ~

so they can be "changeMe" by default

> 2. Application installation is failing, because SSO bundle requires user tables but this bundle was included before UserBundle.

done

> 3. Would be great to have a link in documentation to Google API doc - how to configure Google OAuth.

done

orocrmdeployer:
Hi, few more comments:
# After fail login message like that "User '103552824819952957539' not found." is occurred. It should be standard "Invalid user name or password."
# Please add to doc information about  REDIRECT URIS. ( login/check-google )
# And it would be nice if the sso would be completely optional. Remove google.client_ parameters from parameters.yml and add them to the System Configuration UI

nenadalm:
all fixed

nenadalm:
Updates related with yesterday's skype sent.

* remove use default in the ui
* comma separated values for domains (instead of collection)
* take domains into account for each login

so the only missing thing is the icon in system configuration

nenadalm:
font-awesome used by oro is missing google icon: http://fortawesome.github.io/Font-Awesome/icon/google/

so probably it should be updated to newer version? (also class names are not .icon* but .fa* in newer versions of font-awesome (4)) so icon classes would have to be updated also

nenadalm:
functional tests passed:
https://magecore.atlassian.net/builds/browse/BAP-FUNCTIONALTESTS136-1

PR #227, opened 2015-02-18
BugFix: dropdownButton UI bundle macro

LRotherfield:
Fixes issue with dropdownButton UI bundle macro that caused twig exception when providing the elements key in passed parameters.  

This was due to the looped variable "item" not being passed to dropdownItem but instead the full "parameters" variable.


yurio:
Thanks, @LRotherfield ! Merged to development repository and will be available soon here.

PR #230, opened 2015-02-19
Fix grid caching data issue

### LINKED ISSUE #223: Multiple datagrids with the same name render the data of the first one
In one window there are shown multiple items from one entity. 
The entity has attachments so for each item we display the `attachment-grid` through the `view_content_data_additional_information` placeholder.

Surprisingly (for me) this kind of works in that for each item there is an attachment grid rendered.
The only issue is that the data displayed in all the grids is the data of the first grid.

I investigated this issue and found out that there is a [Backbone collection caching system](https://github.com/orocrm/platform/blob/master/src/Oro/Bundle/DataGridBundle/Resources/public/js/content-manager.js#L21) that uses as a [key for cache the name of the grid](https://github.com/orocrm/platform/blob/master/src/Oro/Bundle/DataGridBundle/Resources/public/js/app/components/datagrid-component.js#L140) (which for all the grids is `attachment-grid`). 

So for my case I overwrite the datagrid-component.js and compose the cache key with the widget id.

Even though I'm not sure that this is an important issue with the datagrid (because the situation I stumbled upon is rare) I still wanted to notify you of it's existence. If you do agree with me and consider it an issue and decide that you want it resolved I can make a pull request with a fix for it.

The fix I propose is to use the grid id as a key for the cache instead of the grid name or concatenate the grid id to the grid name.

dragulceo:
Fixes #223

The only question I had (which was also the reason I did not make this pull request before) was the reasoning behind the datagrid caching. The reasoning is unknown to me but thinking out loud right now I realize that this caching is useful only when you rerender the grid and the data (the collection) is fetched already. So it is safe to use the gridId instead of the gridName (which conflicts with other grids with the same name).

Thank you.

### DIFF
--- a/src/Oro/Bundle/DataGridBundle/Resources/public/js/app/components/datagrid-component.js
+++ b/src/Oro/Bundle/DataGridBundle/Resources/public/js/app/components/datagrid-component.js
@@ -79,6 +79,7 @@ define(function (require) {
         processOptions: function (options) {
             options.$el = $(document.createDocumentFragment());
             options.gridName = options.gridName || options.metadata.options.gridName;
+            options.gridId = options.gridId || options.metadata.options.gridId;
             options.builders = options.builders || [];
             options.builders.push('orodatagrid/js/grid-views-builder');
             options.gridPromise = this.built.promise();
@@ -92,6 +93,7 @@ define(function (require) {
         initDataGrid: function (options) {
             this.$el = options.$el;
             this.gridName = options.gridName;
+            this.gridId = options.gridId;
             this.data = options.data;
             this.metadata = _.defaults(options.metadata, {
                 columns: [],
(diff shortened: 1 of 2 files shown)


akuzmenko-magecore:
To have more control over grid name, you can use `oro_datagrid_build_fullname` twig function. For example:
```
{% set gridName = oro_datagrid_build_fullname('entity-relation-grid', entity_name ~ '-' ~  field_name) %}
```

Grygir:
This solution breaks grid cache functionality. `gridId` is generated each time with unique suffix (e.g. `'grid-accounts-grid-1727381985'`, then `'grid-accounts-grid-464493576'`), so we can not read from the cache by this key.

Regarding your issue https://github.com/orocrm/platform/issues/223, see how grid name is built:
```
    gridName: oro_datagrid_build_fullname(datagrid.name, datagrid.scope),
```
changing datagrid scope you can create similar grids for different entities.

For more details see [DataGrid Scope](https://github.com/orocrm/platform/blob/master/src/Oro/Bundle/DataGridBundle/Resources/doc/backend/scopes.md) documentation.

PR #232, opened 2015-02-25
OroKernel collectBundles fix when bundles.yml empty

nnmer:
When bundles.yml file empty or has no 'bundles:' node - need to filter it correctly. Just ignore such files content.

### DIFF
--- a/src/Oro/Bundle/DistributionBundle/OroKernel.php
+++ b/src/Oro/Bundle/DistributionBundle/OroKernel.php
@@ -134,9 +134,13 @@ abstract class OroKernel extends Kernel
         $exclusions = array();
         foreach ($files as $file) {
             $import  = Yaml::parse($file);
-            $bundles = array_merge($bundles, $this->getBundlesMapping($import['bundles']));
-            if (!empty($import['exclusions'])) {
-                $exclusions = array_merge($exclusions, $this->getBundlesMapping($import['exclusions']));
+            if (null !== $import){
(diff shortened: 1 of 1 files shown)


src/Oro/Bundle/DistributionBundle/OroKernel.php
```diff
@@ -134,9 +134,13 @@ protected function collectBundles()
         $exclusions = array();
         foreach ($files as $file) {
             $import  = Yaml::parse($file);
-            $bundles = array_merge($bundles, $this->getBundlesMapping($import['bundles']));
-            if (!empty($import['exclusions'])) {
-                $exclusions = array_merge($exclusions, $this->getBundlesMapping($import['exclusions']));
+            if (null !== $import){
```
yurio: I think that this check here is redundant because in lines 138 and 141 we have all the necessary checks.

yurio:
@nnmer, thanks for your PR. Please fix comments and after this we can merge your changes.

nnmer:
It is not redundant!

if you do not do that check, you will get an exception:
```
Argument 1 passed to        ::getBundlesMapping() must be of the type array
```

nnmer:
just leave in your bundles.yml
```
bundles:
#      - some bundle
```

and remove the if condition - you will get the exception

yurio:
Thank you for your PR . We accepted it and merged it to dev repository. You will see your changes in public master as soon as possible.

PR #233, opened 2015-02-26
oro:auto-loading-bundles:list command to preview the ordering of bundles...

nnmer:
command to list bundles auto loading and their order

### DIFF
--- a/src/Oro/Bundle/DistributionBundle/Command/ListAutoLoadingBundlesCommand.php
+++ b/src/Oro/Bundle/DistributionBundle/Command/ListAutoLoadingBundlesCommand.php
@@ -1,0 +1,37 @@
+<?php
+namespace Oro\Bundle\DistributionBundle\Command;
+
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Oro\Bundle\DistributionBundle\Console\Grid;
+
+class ListAutoLoadingBundlesCommand extends ContainerAwareCommand
+{
+    protected function configure()
+    {
+        $this
+            ->setName('oro:auto-loading-bundles:list')
+            ->setDescription('List auto loading bundles by their priority');;
+    }
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $bundles = $this->getApplication()->getKernel()->collectBundles();
+
+        if ($bundles) {
+            $output->writeln('<info>Auto loading bundles with priority:</info>');
+
+            $grid = new Grid(3, ['=>', ':']);
+            foreach ($bundles as $bundle) {
+                $grid->addRow(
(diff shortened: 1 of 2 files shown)


hurricane-voronin:
Hi @nnmer, thank you for the PR.
Could you please provide us with information for which cases you need this command.
As an alternative of this command you can see this information in the cache directory: *cache/{environment}/bundles.php*

nnmer:
Hi @hurricane-voronin  .
The command will show me the order of bundles together with the priority that was set to them. So i can easily see what priority i need to set to new bundle if i need to load it in specific order. Using 
*cache/{environment}/bundles.php*
I cannot do what I described above.

iklymenko:
@nnmer, thanks for the explanation! I have additional question. In the  command output there is third column in the grid that shows the bundles which requires Kernel object in the bundle class. How this information could be useful? 
Also, I want to note that it is not obvious for the user what kind information this column shows.

src/Oro/Bundle/DistributionBundle/Command/ListAutoLoadingBundlesCommand.php
```diff
@@ -0,0 +1,37 @@
...
+
+class ListAutoLoadingBundlesCommand extends ContainerAwareCommand
+{
+    protected function configure()
+    {
+        $this
+            ->setName('oro:auto-loading-bundles:list')
+            ->setDescription('List auto loading bundles by their priority');;
+    }
+    protected function execute(InputInterface $input, OutputInterface $output)
```
iklymenko: Missing import of the Symfony\Component\Console\Output\OutputInterface namespace

src/Oro/Bundle/DistributionBundle/Command/ListAutoLoadingBundlesCommand.php
```diff
@@ -0,0 +1,37 @@
+<?php
+namespace Oro\Bundle\DistributionBundle\Command;
+
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputArgument;
```
iklymenko: This namespace is unused, so could be deleted.

src/Oro/Bundle/DistributionBundle/OroKernel.php
```diff
@@ -121,7 +121,7 @@ function (\SplFileInfo $current) use (&$paths) {
     /**
      * @return array
      */
-    protected function collectBundles()
+    public function collectBundles()
```
yshyshkin: Instead of making this method public (that is not really good), please, move search and collection logic to separate class, and reuse it in both cases without making method collectBundles() public. Class can be called e.g. Oro\Bundle\DistributionBundle\Bundle\Collector.

orocla:

nnmer:
ignore this PR, i will not do any contributions in future.

DimaSoroka:
Thanks @nnmer, I hope OroPlatform was helpful for you and you'll change your mind in future.

PR #238, opened 2015-03-10
Bug: Datagrids return the wrong number of rows when paginating a complex joined query

LRotherfield:
Fix: For non grouped queries, paginate results using Doctrine Paginator methodolgy;
  Perform a Count query using DISTINCT keyword.
  Perform a Limit Subquery with DISTINCT to find all ids of the entity in from on the current page.
  Perform a WHERE IN query to get all results for the current page.

Includes updated test scenarios for CountQueryBuilderOptimizer.

I am happy to make any updates required, this works for all of the scenarios I have tested

### DIFF
--- a/src/Oro/Bundle/BatchBundle/ORM/QueryBuilder/CountQueryBuilderOptimizer.php
+++ b/src/Oro/Bundle/BatchBundle/ORM/QueryBuilder/CountQueryBuilderOptimizer.php
@@ -95,7 +95,7 @@ class CountQueryBuilderOptimizer
             $this->addJoins($qb, $parts);
         }
         if (!$parts['groupBy']) {
-            $fieldsToSelect[] = $this->getFieldFQN($this->idFieldName);
+            $fieldsToSelect[] = sprintf('DISTINCT %s', $this->getIdFieldFQN());
         }
 
         if ($parts['where']) {
@@ -106,6 +106,16 @@ class CountQueryBuilderOptimizer
         $this->qbTools->fixUnusedParameters($qb);
 
         return $qb;
+    }
+
+    /**
+     * Get Identifier field's fully qualified namespace
+     *
+     * @return string
+     */
+    public function getIdFieldFQN()
+    {
+        return $this->getFieldFQN($this->idFieldName);
     }
 
     /**
--- a/src/Oro/Bundle/BatchBundle/ORM/QueryBuilder/CountQueryBuilderOptimizer.php
+++ b/src/Oro/Bundle/BatchBundle/ORM/QueryBuilder/CountQueryBuilderOptimizer.php
@@ -52,7 +52,7 @@ class CountQueryBuilderOptimizer
      * @param QueryBuilder $originalQb
      * @return QueryBuilder
      */
-    public function getCountQueryBuilder(QueryBuilder $originalQb)
+    public function getCountQueryBuilder(QueryBuilder $originalQb, $addOrderBy = false)
     {
         $this->setOriginalQueryBuilder($originalQb);
         $parts = $this->originalQb->getDQLParts();
@@ -75,7 +75,7 @@ class CountQueryBuilderOptimizer
             foreach ($groupByFields as $key => $groupByField) {
                 $alias = '_groupByPart' . $key;
                 $usedGroupByAliases[] = $alias;
-                $fieldsToSelect[] = $groupByField . ' as ' . $alias;
+                $fieldsToSelect[$groupByField] = $groupByField . ' as ' . $alias;
             }
             $qb->groupBy(implode(', ', $usedGroupByAliases));
         } elseif (!$parts['where'] && $parts['having']) {
@@ -94,8 +94,22 @@ class CountQueryBuilderOptimizer
         if ($parts['join']) {
             $this->addJoins($qb, $parts);
         }
+
         if (!$parts['groupBy']) {
-            $fieldsToSelect[] = sprintf('DISTINCT %s', $this->getIdFieldFQN());
+            $qb->distinct(true);
+            $fieldsToSelect[$this->getIdFieldFQN()] = $this->getIdFieldFQN();
+        }
+
+        if ($addOrderBy && $parts['orderBy']) {
+            foreach ($parts['orderBy'] as $orderExpr) {
+                $orderParts = explode(' ', $orderExpr);
+                $field = $this->qbTools->getFieldByAlias($orderParts[0]);
+                if ($field && !isset($fieldsToSelect[$orderParts[0]])) {
+                    // This is an aliased field so add it to the select
+                    $fieldsToSelect[$orderParts[0]] = sprintf('%s AS %s', $field, $orderParts[0]);
+                }
+                $qb->addOrderBy($orderExpr);
+            }
         }
 
         if ($parts['where']) {
--- a/src/Oro/Bundle/DataGridBundle/Extension/Pager/Orm/Pager.php
+++ b/src/Oro/Bundle/DataGridBundle/Extension/Pager/Orm/Pager.php
@@ -4,6 +4,7 @@ namespace Oro\Bundle\DataGridBundle\Extension\Pager\Orm;
 
 use Doctrine\ORM\Query;
 use Doctrine\ORM\QueryBuilder;
+
 use Oro\Bundle\BatchBundle\ORM\Query\QueryCountCalculator;
 use Oro\Bundle\BatchBundle\ORM\QueryBuilder\CountQueryBuilderOptimizer;
 use Oro\Bundle\DataGridBundle\Extension\Pager\PagerInterface;
@@ -138,18 +139,16 @@ class Pager extends AbstractPager implements PagerInterface
 
         /** @var QueryBuilder $query */
         $query = $this->getQueryBuilder();
+        if (count($this->getParameters()) > 0) {
+            $query->setParameters($this->getParameters());
+        }
         $countQb = $this->countQueryBuilderOptimizer->getCountQueryBuilder($this->getQueryBuilder());
 
         $query->setFirstResult(null);
         $query->setMaxResults(null);
 
-
-        if (count($this->getParameters()) > 0) {
-            $query->setParameters($this->getParameters());
-        }
-
         if (!$countQb->getDQLPart('groupBy')) {
-            $this->setPaginiationData($countQb);
+            $this->setPaginationData($countQb);
             foreach ($countQb->getQuery()->getResult() as $data) {
                 $inArray[] = current($data);
             }
@@ -159,7 +158,7 @@ class Pager extends AbstractPager implements PagerInterface
                 return;
             }
         }
-        $this->setPaginiationData($query);
+        $this->setPaginationData($query);
     }
 
     /**
@@ -250,7 +249,7 @@ class Pager extends AbstractPager implements PagerInterface
     /**
      * @param QueryBuilder $query
      */
-    private function setPaginiationData(QueryBuilder $query)
+    private function setPaginationData(QueryBuilder $query)
     {
         if (0 == $this->getPage() || 0 == $this->getMaxPerPage() || 0 == $this->getNbResults()) {
             $this->setLastPage(0);
--- a/src/Oro/Bundle/DataGridBundle/Extension/Pager/Orm/Pager.php
+++ b/src/Oro/Bundle/DataGridBundle/Extension/Pager/Orm/Pager.php
@@ -4,7 +4,6 @@ namespace Oro\Bundle\DataGridBundle\Extension\Pager\Orm;
 
 use Doctrine\ORM\Query;
 use Doctrine\ORM\QueryBuilder;
-
 use Oro\Bundle\BatchBundle\ORM\Query\QueryCountCalculator;
 use Oro\Bundle\BatchBundle\ORM\QueryBuilder\CountQueryBuilderOptimizer;
 use Oro\Bundle\DataGridBundle\Extension\Pager\PagerInterface;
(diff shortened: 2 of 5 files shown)


LRotherfield:
I am happy to make any updated required, this works for all of the scenarios I have tested

src/Oro/Bundle/BatchBundle/ORM/QueryBuilder/CountQueryBuilderOptimizer.php
```diff
@@ -95,7 +95,7 @@ public function getCountQueryBuilder(QueryBuilder $originalQb)
             $this->addJoins($qb, $parts);
         }
         if (!$parts['groupBy']) {
-            $fieldsToSelect[] = $this->getFieldFQN($this->idFieldName);
+            $fieldsToSelect[] = sprintf('DISTINCT %s', $this->getIdFieldFQN());
```
hugeval: I suppose here can come cross join or table with composed primary key, so it looks like we shouldn't use distinct keyword

src/Oro/Bundle/DataGridBundle/Extension/Pager/Orm/Pager.php
```diff
@@ -4,7 +4,6 @@
 
 use Doctrine\ORM\Query;
 use Doctrine\ORM\QueryBuilder;
-
```
hugeval: This line separates vendors and Oro namespaces, maybe it should stay here

src/Oro/Bundle/DataGridBundle/Extension/Pager/Orm/Pager.php
```diff
@@ -139,24 +138,28 @@ public function init()
 
         /** @var QueryBuilder $query */
         $query = $this->getQueryBuilder();
+        $countQb = $this->countQueryBuilderOptimizer->getCountQueryBuilder($this->getQueryBuilder());
 
         $query->setFirstResult(null);
         $query->setMaxResults(null);
 
+
```
hugeval: I believe it is necessary to remove new line because of too much empty lines

src/Oro/Bundle/DataGridBundle/Extension/Pager/Orm/Pager.php
```diff
@@ -139,24 +138,28 @@ public function init()
...
         if (count($this->getParameters()) > 0) {
             $query->setParameters($this->getParameters());
```
hugeval: Pager can receive additional parameters, it looks like $countQb should consider them too

src/Oro/Bundle/DataGridBundle/Extension/Pager/Orm/Pager.php
```diff
@@ -243,4 +246,21 @@ protected function retrieveObject($offset)
 
         return $results[0];
     }
+
+    /**
+     * @param QueryBuilder $query
+     */
+    private function setPaginiationData(QueryBuilder $query)
```
hugeval: Maybe 'setPaginiationData' is typo and should becomes 'setPaginationData'

hugeval:
Hello @LRotherfield, thank you for great pull request. Please consider my notes. Can you provide several scenarios those you tested?

src/Oro/Bundle/BatchBundle/ORM/QueryBuilder/CountQueryBuilderOptimizer.php
LRotherfield: I think I may be missing something as I don't see how this value could be a join or composite primary key because [here](https://github.com/LRotherfield/platform/blob/hotfix/datagrid-complex-join-pagination/src/Oro/Bundle/BatchBundle/ORM/QueryBuilder/CountQueryBuilderOptimizer.php#L185) you are making sure that only one field is returned as the primary id field?

If there is an issue then this code wont work because DISTINCT is key to this patch https://github.com/doctrine/doctrine2/blob/a41e774bb1f84d6f139438e40c45d54f042978c0/tests/Doctrine/Tests/ORM/Tools/Pagination/LimitSubqueryWalkerTest.php

src/Oro/Bundle/DataGridBundle/Extension/Pager/Orm/Pager.php
LRotherfield: No problem, I will put this back

src/Oro/Bundle/DataGridBundle/Extension/Pager/Orm/Pager.php
LRotherfield: No problem, I will remove this

src/Oro/Bundle/DataGridBundle/Extension/Pager/Orm/Pager.php
LRotherfield: Yes typo, will sort this out

LRotherfield:
The scenario's I covered were:
* Do oro data grids work still, following tested using filters, pagination and sorters:
  * User grid
  * Custom entity grid
  * Email templates grid
* Complex join pagination (14 entities in grid but complex join returns 23 rows):
  * Limit number or results to 10 per page (with fix returns 10 results, without returns 6 results which is 10 rows from db)
  * Limit number of results to 10 per page and navigate through pages (with fix, 2 pages, second page has 4 results, without fix has 3 pages and second page has ~6 results which again maps to 10 rows from db)
  * Apply filters to return only 11 results and use pagination (with fix, 2 pages, 10 and 1 results respectively, without fix still odd number of results)
  * Change bound parameter (our grids show data bound to a specific locale so switch locale) with fix, results are as expected, no unexpected results returned

These are the scenarios that our datagrids get subject to and thus those I have tested.  If you have any others then please let me know.

src/Oro/Bundle/DataGridBundle/Extension/Pager/Orm/Pager.php
LRotherfield: Are you happy for me to bind them to query before line [141](https://github.com/LRotherfield/platform/blob/hotfix/datagrid-complex-join-pagination/src/Oro/Bundle/DataGridBundle/Extension/Pager/Orm/Pager.php#L141) so that the clone on line [60](https://github.com/LRotherfield/platform/blob/hotfix/datagrid-complex-join-pagination/src/Oro/Bundle/BatchBundle/ORM/QueryBuilder/CountQueryBuilderOptimizer.php#L60) of the countQueryBuilder gets them?

LRotherfield:
@hugeval I have pushed some changes that I can rebase if this patch is accepted.  Can you confirm if the setParameters issue has been solved?  Also need to understand if using DISTINCT where I have added it is invalid or not?

LRotherfield:
Looks like sorting with paginated results does not work properly with this fix, as fields are not ordered in count optimised query.  Trying to come up with solution, will commit when done

LRotherfield:
@hugeval, @DimaSoroka what is the reason for the sorter extension running after the pager extension?  I need the order by queries to make this patch work but I cannot get them if with the current extension priorities and I am just wondering if there is a technical reason to have these two extensions run in this order?

akuzmenko-magecore:
Hi, @LRotherfield.
I tried to reproduce the issue with getting wrong number of rows from datagrids using your scenarios and I could not. Is this issue reproducible on standard OroCRM installation or it appears after some customization? Could you, please, provide more details about the testing process you are using?

As to answer your question about DISTINCT being viable, I think it is. As the matter of fact, we used to have it there and it has been removed in:
https://github.com/orocrm/platform/commit/7688609e2e43a1ec2f2c32b67ce5672f93765b97#diff-38268d37f41144234fa2b1fe42221e02
By the way, a small note here. It's better to use "$qb->distinct(true);" to add DISTINCT statement to select.
Also I'd recommend to not use "private" methods since we often have issues with them when extending functionality later on.
About the order of Sorter and Pager extensions. I don't think there is any particular reason the sorter is running after the pager. But I don't think that pager should depend on the sorter.

LRotherfield:
Hi @hugeval,

Thanks for the responses:
>Is this issue reproducible on standard OroCRM installation or it appears after some customisation? Could you, please, provide more details about the testing process you are using?

We are not using customisation on any of the datagrid components, however the entities are our own.
The entity that has an issue has the following mapping:
`Entity -<one to many>- translations -<one to many>- versions`

When we fetch all of the translations and versions for the entity and use them in the grid, then we get the odd results.

>As to answer your question about DISTINCT being viable, I think it is. As the matter of fact, we used to have it there and it has been removed in

I actually was reading through this commit yesterday and I noticed you used the $qb->distinct(true) [where true was dependant on hasJoins].  I can fix that in my commit

>Also I'd recommend to not use "private" methods since we often have issues with them when extending functionality later on.

I can make this protected

>About the order of Sorter and Pager extensions. I don't think there is any particular reason the sorter is running after the pager. But I don't think that pager should depend on the sorter.

I have a fix for the pager with regards to order by, but due to the nature of [doctrine's pagination solution](http://doctrine-orm.readthedocs.org/en/latest/tutorials/pagination.html) (creating a where in statement) I need to make sure that all parts of the query have been added before running the pagination query so that part's like order by have been added.  Does that make sense?

LRotherfield:
@hugeval, @akuzmenko-magecore Hi guys, I pushed my fixes for the order by issue, could you please have a review and let me know what you think

LRotherfield:
Hi Guys

Any feedback on this?  If you are not keen, that is fine, we needed the pagination fix so I have implemented it using overrides in our environment. If you need me to do anymore work on it, please let me know.

Kind Regards
Luke

akuzmenko-magecore:
Hi Luke,
Thank you for being patient.
We've been able to reproduce this (or similar) issue on reports. Although it seems that the main cause of the issue is not in incorrect pagination but in displaying wrong records. The count is calculated correctly but not all of the records are being shown. 
We'll need to have a deeper investigation. It seems that using "DISTINCT" will not solve the problem entirely and can have a harmful effect on overall grid performance. It's an important issue and we are going to work on it as soon as we can.

LRotherfield:
Hi @akuzmenko-magecore,

Thanks for getting back to me.  I am sure that this is the same pagination issue as Doctrine themselves had which required the DISTINCT query to resolve.  I will try and get a bundle mocked up with entities that shows this for you in a vanilla install of the platform.  I will update here when I have put it together.

akuzmenko-magecore:
Hi Luke,
We've made some small tweaks to Data Grids in our new 1.7 release. Possibly they can solve your problem.

orocla:

LRotherfield:
Hi Guys

Sorry about such a long wait on me to feedback.  My implementation does not work for mass actions so is not valid.  Our project is currently running version 1.6 as it is in production and 1.7 caused a break in one of our modules that we need to upgrade.  However I am in the process of upgrading to version 1.9 of the platform after which I will test our use case and see if the datagrid now works.

LRotherfield:
I have upgraded to 1.9 (although I still have much to fix) and can confirm that my grid is getting totals wrong and displaying the wrong number of children due to the relationships of those entities.  I will try and get a vanilla version setup as an example (as I promised months ago) so you can see if it is just a mis-configuration on my part or really a bug

LRotherfield:
Hi guys

As I promised, here is an example bundle that will show the error I am getting, please let me know if I am just doing something silly.

https://github.com/LRotherfield/oro_datagird_count_error

Have a wonderful Christmas

DimaSoroka:
Thanks! Merry Christmas! We'll take a look into your example shortly.

x86demon:
Hi, @LRotherfield 
May you backmerge current master? DISTINCT is already added to COUNT query in case when main query contains it for identifier.

Please, check latest changes with your example.

LRotherfield:
Hi @x86demon 
I tested this module on master today, I just cloned master and built the bundle then committed it up.

LRotherfield:
Here is the bug in action showing 4 results but reporting 11
screen shot 2015-12-24 at 20 08 14
Here is the system information, as I said, cloned today
screen shot 2015-12-24 at 20 08 02

x86demon:
@LRotherfield I've checked your example on current master. Example was modified - DISTINCT on Primary Key was added to trigger DISTINCT adding to COUNT query. 

```yaml
select:
    - DISTINCT i.id
    - i
```
screenshot
Screenshot taken on current master with modified grid select.

DISTINCT could not be always applied, as it may affect items count for grids where left joins are used and records are not hydrated like in given example.

```yaml
defacto-item-grid-no-hydration:
        extended_entity_name: %defacto.example.entity.item.class%
        options:
            entityHint: Item
        source:
            type: orm
            query:
                select:
                    - t.locale
                    - v.name
                from:
                    - { table: %defacto.example.entity.item.class%, alias: i }
                join:
                    left:
                        - { join: i.translations, alias: t }
                        - { join: t.version, alias: v }
        columns:
            locale:
                data_name: locale
            version:
                data_name: name
```

screenshot2
Screenshot is taken in case when DISTINCT is always enabled.

Please, note that CountQueryBuilderOptimizer in your branch differs from CountQueryBuilderOptimizer in master branch.
https://github.com/orocrm/platform/blob/master/src/Oro/Bundle/BatchBundle/ORM/QueryBuilder/CountQueryBuilderOptimizer.php
https://github.com/LRotherfield/platform/blob/hotfix/datagrid-complex-join-pagination/src/Oro/Bundle/BatchBundle/ORM/QueryBuilder/CountQueryBuilderOptimizer.php

LRotherfield:
Brilliant, thanks @x86demon I will test this out on our actual install but I think it will work in which case I will close this ticket.  As you say my code is completely out of date, I think I wrote it for platform version 1.6 :)

Thanks for a response on Christmas, have a great one

Luke

iklymenko:
Hi Luke,
did you had a chance to test it?

LRotherfield:
@iklymenko yes I did thanks, it did not work but I don't think it is due to an issue with this code any longer so I will close this pull request, thanks very much to your team for the feedback and help

PR #249, opened 2015-04-07
Fixed crop mode for image thumbnails

ramunasd:


akuzmenko-magecore:
Hi @ramunasd,
I don't think that inset mode should be used by default. I can imagine that it is preferable for some cases. But this would make user avatars look inconsistent - different shapes and sizes for different users.
Why do you consider that images should be cropped this way, do you have any particular case?
If you like, you can prepare a pull request to make this option configurable somehow.

ramunasd:
@akuzmenko-magecore If this controller is used for avatars, do you don't think it should be in UserBundle, not in common AttachmentBundle. I mean special cases should look like special. Avatars is special cases.

Configurable filters would be great.

akuzmenko-magecore:
User avatars was just an example. The same controller used in Contact avatars and I think this should be fine for other entities as well. The idea is to keep the size of the images the same. Imagine displaying several entities with images on the same page.
If you have a case where this kind of crop is really needed, you are welcome to share it with us and will consider if we can make this improvement.

ramunasd:
@akuzmenko-magecore actually, this is platform and there can be many use cases, depending on what is primary application purpose :) Do i must list all possible cases with wrong cropping settings?

DimaSoroka:
@ramunasd, please share with us your particular case. As proposed solution is not general enough we would like to understand your requirements better and see how we can cover general case on th platform level. If this PR is only about avatars cropping mode, this could be done as configuration option.

ramunasd:
@DimaSoroka yeah, PR is not general enough, i would like to make correct one. Configurable option would be great for me.

I'm closing this PR because of incorrect solution and misunderstanding between me and project developers.

PR #255, opened 2015-04-23
Extend exception in AssertAssembler to give more info

sbani:
Added property_path from option to see where the option is missing

If property_path is always there, I will add it to unittest and and change exception with one line of code.


akuzmenko-magecore:
As I understand, `property_path` makes sense only for AttributeAssembler in WorkflowBundle. I think, it would be better to customize implementation of `assertOptions()` method there. Keep in mind that it can be empty in some cases.

PR #258, opened 2015-05-07
Check whether the cron is enabled or not

sbani:
If cron is not enabled, skip it
This is useful if you need to disable a cronjob for specific reason
Example: Disable cron for env dev


sbani:
Is already implemented. Made a wring check. Sorry

PR #270, opened 2015-06-08
Require PHP 5.4

pierredup:
The CronBundle uses short-array syntax a lot, which isn't compatible with PHP 5.3


DimaSoroka:
Changes were applied in scope of other commit, closing this PR.

PR #282, opened 2015-06-24
Replace iconv(..., 'ascii//TRANSLIT', ...) with Intl Transliterator

turboezh:
Hi.
I've faced to a problem with iconv suddenly, when tried to add options with Russian letters to a select field.
> PHP message: PHP Notice:  iconv(): Detected an illegal character in input string in /Users/alex/www/orocrm/www/orocrm/vendor/oro/platform/src/Oro/Bundle/EntityExtendBundle/Tools/ExtendHelper.php on line 173

May be it's a bug, may be it depends on OS environment or PHP version. I didn't investigate source of the problem and just have replaced iconv with Intl Transliterator. I think it's more reliable way to transliterate strings.


akuzmenko-magecore:
Hi @turboezh,
We've accepted your pull request and it will be available in future releases. Thank you for your contribution.

DimaSoroka:
Closing this PR as changes already merged.

PR #289, opened 2015-07-17
Upgraded version of Nelmio API doc bundle

AshleyDawson:
We require a feature in version 2.9 that allows us to segregate our domain-specific API documentation from the ORO API


iklymenko:
Hi Ashley,
changes were applied in our latest release. Thank you for the contribution.

PR #291, opened 2015-07-23
Return $this when calling setReminderData

sbani:
Return the Reminder object

Just to make sure all setter return $this


sbani:
Probably one of the smallest PR ever

iklymenko:
Hi @sbani,
size doesn't matter, thanks for your contribution! :)
We applied your suggested changes, so soon they will be available here.

PR #293, opened 2015-07-24
Fix 'null' ID in route error caused when a PHP-based form validation …

nmallare:
…constraint is fired.

Use Case
 * Base Entity is extended using ExtendExtension to add collection (many-to-many)
 * User creates custom validation constraint to make sure the same item isn't added to `oro_collection` twice
 * Form is created adding only new ExtendExtension field
 * Form is submitted without triggering the validation and data is persisted to the database
 * Form is re-submitted and validation constraint fires
 * Observe `Parameter "id" for route "oro_entity_detailed" must match "[^/]++" ("" given) to generate a corresponding URL` Exception

Please reach out with comments or questions.


hurricane-voronin:
Hi @nmallare,
Thank you for the PR.
Changes already merged.

nmallare:
@hurricane-voronin Should I close this PR, then?

hurricane-voronin:
@nmallare , Yes. Thanks.

PR #297, opened 2015-07-30
Fix create extend field form after many-to-one relation

ishakuta:
STR:
- create extend one-to-many relation from Contact to ContactAddress
- update schema
- go to Contact entity and try to add another extend field

Expected: form to fill the entity field name and select type
Actual: 500 error "A model for "OroCRM\Bundle\ContactBundle\Entity\ContactAddress::contact_addresses" was not found (500 Internal Server Error)"


hurricane-voronin:
Hi @ishakuta,
Thank you for the PR.

We have already fixed this issue.

PR #305, opened 2015-08-18
fixing the value of `$isAssignGranted`

2ndkauboy:
I was wondering how to hide the ownershipo dropdown for an entity with the access level set to `user`, so I asked in the forum: http://www.orocrm.com/forums/topic/how-do-i-remove-the-owner-dropdown

As explained, I set the assign permission to `none`. When I opened an entity item for update, the ownershio dropdown was gone. But when I created an entity item, I could still change set the owner.

I found out, that when creating an entity item, the `OwnerFormExtension` is checking, if the user has the `CREATE` permission. If this is true, the flag `$isAssignGranted` is set to true.

This makes no sense, as it should be possible to rescrict a user not being able to assign an entity item, even on creation. When the dropdown is removed, the item will be assigned to the user, just as someone would expect.

Is there any reason, which I might not see, why the original code is checking against the `CREATE` permission?

### DIFF
--- a/src/Oro/Bundle/OrganizationBundle/Form/Extension/OwnerFormExtension.php
+++ b/src/Oro/Bundle/OrganizationBundle/Form/Extension/OwnerFormExtension.php
@@ -151,7 +151,7 @@ class OwnerFormExtension extends AbstractTypeExtension
 
         $this->fieldName = $metadata->getOwnerFieldName();
 
-        $this->checkIsGranted('CREATE', 'entity:' . $dataClassName);
+        $this->checkIsGranted('ASSIGN', 'entity:' . $dataClassName);
         $defaultOwner = null;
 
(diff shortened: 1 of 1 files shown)


yurio:
Hi, @2ndkauboy.
If you want to hide user select box in case if your user have User access level to create, you should rewrite addUserOwnerField method of OwnerFormExtension class in your project. 
So, your addUserOwnerField method should looks something like:

```
 /**
  * {@inheritdoc}
  */
protected function addUserOwnerField($builder, $dataClass, $permission = "CREATE", $data = null, $entityId = 0)
    {
             if ($this->accessLevel !== 'USER') {
                  parent::addUserOwnerField($builder, $dataClass, $permission, $data, $entityId);
             }
}
```

Change Create permission to Assign is not a good idea because Assign permission was created to be able to limit users list during edit some entities.
When you create some record, Create permission should limit this list.

PR #326, opened 2015-10-15
fix in UIBundle:view.html.twig

apoberez:
important when entity doesn't have "createdAt" or "updatedAt"


DimaSoroka:
thanks @apoberez, just for testing purposes could you please describe how to reproduce the issue this PR is fixing

apoberez:
if I have new entity without fields createdAt|updatedAt or they have other names such as "created" view will fail to render this entity

apoberez:
we can define some interface like "Timestampable" for example to unify this across hole project

iklymenko:
Hi @apoberez,
your changes were successfully applied using patch of current PR because the source repository is unknown. Thanks for your contribution!

PR #327, opened 2015-10-18
don't use empty array as IN() parameters in DQL

xabbuh:
When the `UserCalendarProvider::getCalendarDefaultValues()` method is
called from the `CalendarManager::getCalendars()` method, an empty array
can be passed. However, using an empty array as a value of the DQL
`IN()` clause leads to a syntax error. Thus, we can prematurely return
here with an empty result.


yurio:
Hi, @xabbuh.
Thanks for the pull request. 
Changes were applied in our latest release. Thank you for the contribution.

DimaSoroka:
Closing this PR as fix was applied

PR #329, opened 2015-10-20
Adding validation to avoid overloading in flush process working with …

cesar-giraldo:
…many workflows at same time.

Renaming some vars.


alex-n-2k7:
thanks a lot, we have processed your PR, soon it will be merged into public

src/Oro/Bundle/WorkflowBundle/Model/WorkflowManager.php
```diff
@@ -17,6 +17,11 @@
 class WorkflowManager
 {
     /**
+     * Limit of workflows items to make a flush 
+     */
+    const LIMIT_WORKFLOWS_TO_PROCESS = 50;
```
ReenExe: why now use array_chunk?

PR #333, opened 2015-10-29
fix not required on js type validation

Devolicious:
This fixes the js validation for integer type when it's not a required field.


Grygir:
Thank you for your contribution. Changes are merged into dev repo. With next update of orocrm/platform they get into public.

PR #353, opened 2015-11-23
Add helper methods to grid configuration

ishakuta:
provide a way to manipulate grid configuration without knowing exact inner array structure
- add/remove column, filter, sorter
- add mass action
- update label
- join table
- add enum filter
- add twig column

cc @vsoroka @AlexandrDmitriev please let me know if you find it useful, or have any ideas about it, thanks.

### DIFF
--- a/src/Oro/Bundle/DataGridBundle/Datagrid/Common/DatagridConfiguration.php
+++ b/src/Oro/Bundle/DataGridBundle/Datagrid/Common/DatagridConfiguration.php
@@ -2,8 +2,259 @@
 
 namespace Oro\Bundle\DataGridBundle\Datagrid\Common;
 
+use Doctrine\ORM\EntityRepository;
+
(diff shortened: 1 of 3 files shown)


alsma:
Hi! Not sure that it fits in initial idea of configuration object. As far as I remember it also user in some other parts of platform, so code you added is not suitable for system wide usage. 

Probably it should be separate helper object or static utils... let's say, most methods you wrote are related to functionality added by extensions(for instance enums) consequently it should not be part of general config object IMO.

cc @vsoroka

orocla:

AlexandrDmitriev:
Hi, I agree with Alexander Smaga, but as for me we can merge it because I am not sure we will implement more suitable implementation in nearest future. Also as for me, when we will create appropriate helper will be easy to replace all usages of this DatagridConfiguration methods.

AlexandrDmitriev:
Merged

PR #367, opened 2015-12-17
update service deprecated

dimitriseguin:
transform service security_context to security.token_storage

### DIFF
--- a/src/Oro/Bundle/DashboardBundle/Resources/config/services.yml
+++ b/src/Oro/Bundle/DashboardBundle/Resources/config/services.yml
@@ -72,7 +72,7 @@ services:
             - @oro_dashboard.factory
             - @doctrine.orm.entity_manager
             - @oro_security.acl_helper
-            - @security.context
+            - @security.token_storage
 
     oro_dashboard.factory:
         class: %oro_dashboard.factory.class%


orocla:

mbessolov:
This change breaks backwards compatibility.

As the Oro\Bundle\DashboardBundle\Model\Manager class is not _final_, and the $securityContext property is _protected_, it is possible that 3-rd party developers extend the Manager class, and use its $securityContext property.

If they rely on $securityContext to be an instance of [\Symfony\Component\Security\Core\SecurityContext](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/Security/Core/SecurityContext.php) and are calling its method [**isGranted()**](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/Security/Core/SecurityContext.php#L100-L103), and we pass an instance of [\Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorage](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/Security/Core/Authentication/Token/Storage/TokenStorage.php) instead, which does not have such a method, a user will simply get a fatal error `PHP Fatal error: Call to undefined method TokenStorage::isGranted() in ...`

PR #377, opened 2016-01-07
Widget grid configuration

ishakuta:
- allow to pass widget configuration params into grid widget

### DIFF
--- a/src/Oro/Bundle/DashboardBundle/Controller/DashboardController.php
+++ b/src/Oro/Bundle/DashboardBundle/Controller/DashboardController.php
@@ -320,14 +320,16 @@ class DashboardController extends Controller
      * )
      * @Template("OroDashboardBundle:Dashboard:grid.html.twig")
      *
-     * @param string $gridName
+     * @param string  $widget
+     * @param string  $gridName
+     * @param Request $request
      *
      * @return array
      */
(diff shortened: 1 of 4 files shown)


ishakuta:
please let me know if you'll find it useful, so I'll add docs and tests

artml:
@ishakuta what is the perceived use case for this? Make grid widgets like "recent contacts" configurable?

ishakuta:
@artml yes, right it's possible to change title, 
imagine this use case: user wants to configure some value to bind it to the datagrid parameter, something like date range for the "Opportunities By Lead Source" (chart) widget, but for the grid.

The problem here is that master branch contains [hardcoded empty params](https://github.com/orocrm/platform/blob/master/src/Oro/Bundle/DashboardBundle/EventListener/WidgetConfigurationLoadListener.php#L42) array that blocks such use case.
Maybe there's another way to achieve that, but it looks like a bug to me, imho.

hurricane-voronin:
Hi @ishakuta ,
>please let me know if you'll find it useful, so I'll add docs and tests

Will be great if you add it.

hurricane-voronin:
Hi @ishakuta, 
Thank you for the PR.  It was merged to development repository, so you should see you commit in public repository soon. It will be great if you add docs/tests for it in the new PR.

PR #380, opened 2016-01-12
Fixed very important bug in BufferedQueryResultIterator

stalxed:
Hello Oro team!

I found bug in BufferedQueryResultIterator.

For example, we have 100000 records for some query(in our case $query).
And we use this code:

``` php
$iterator = new BufferedQueryResultIterator($query);
$iterator->setBufferSize(100);
$iterator->rewind();
$iterator->rewind();
foreach ($iterator as $record) {
    echo $record->getId() . PHP_EOL;
}
```

This code displays the identifiers for 100100 items(without this fix of course).
Why 100100? Because first 100 rows(=buffer size) will be iterated twice.

Please include this fix for version 1.9.x!!!

### DIFF
--- a/src/Oro/Bundle/BatchBundle/ORM/Query/BufferedQueryResultIterator.php
+++ b/src/Oro/Bundle/BatchBundle/ORM/Query/BufferedQueryResultIterator.php
@@ -239,6 +239,8 @@ class BufferedQueryResultIterator implements \Iterator, \Countable
         $this->position   = -1;
         $this->current    = null;
 
+        unset($this->rows);
+
         $this->next();
     }
 


stalxed:
I registered here http://www.orocrm.com/contributor-license-agreement
> Oro, Inc. CLA Signature
> Your request has been submitted. Please allow for some time for your submission to be processed.

orocla:

alex-n-2k7:
Hi @stalxed 

I've been reviewing your PR and couldn't reproduce the issue in the same way as you described.
Without calling rewind() all works fine, actually rewind will be called automatically on loop starts. 
Actually it's not clear why do you need rewind before loop, especially twice ?

But I noticed another issue - with calling rewind method I can get only first page (bufferSize) of records, e.g. in any Controller:

```php
   ...
   use Oro\Bundle\BatchBundle\ORM\Query\BufferedQueryResultIterator;
   ...

    public function indexAction()
    {
        $em = $this->container->get('doctrine')->getManager();
        $source = $em->createQueryBuilder()
            ->select('o')
            ->from('OroUserBundle:User', 'o');

        $iterator = new BufferedQueryResultIterator($source);
        $iterator->setBufferSize(10);
        $iterator->rewind();
        $iterator->rewind();

        $count = 0;
        foreach ($iterator as $record) {
            var_dump($record->getId());
            $count++;
        }
        var_dump($count);
        die();
        ...
```
In this case if you have 100 records  you will get only first 10 (without your's fix).
And, unfortunately, it gives 0 records with your's fix, so we cannot apply it.

Also, it will be great if you can provide unit test(s) which will cover your case.

DimaSoroka:
@stalxed could you please help with addition information per @alex-n-2k7 comment?

stalxed:
@DimaSoroka , @alex-n-2k7 I will test this issue this week.

iklymenko:
@stalxed hi, do you have any news about regarding the issue? Thanks!

mkudelya:
Close pull request because it has no activity. Feel free open it if you have any question.

PR #396, opened 2016-02-10
Remove pageSize strict type checking in DataGridBundle

### LINKED ISSUE #395: Non-default "Items Per Page" hangs data grid when changing grid view
In System --> Configuration --> General setup --> Display settings, changing the "Items Per Page By Default" option under Data Grid Settings causes the data grid to hang when changing grid views. To duplicate:

1. Change the setting mentioned above to something non-default (say, 100).
2. Browse to a data grid (say, Sales --> Leads).
3. Click the "All Leads" grid view to something else (create one first, if needed).
4. Watch as the new leads are loaded and displayed, but the progress spinner never quits and the View Per Page remains greyed out. 

At this point, checking the console reveals an error message along the lines of:

```TypeError: currentSizeLabel[0] is undefined```

Returning the option mentioned above to the default value fixes the data grid display.

### PR DESCRIPTION
**Pull request by cmenning**
Fixes #395.

Non-default configurations for are provided as strings, and therefore fail the strict comparison. This leaves no elements in the currentSizeLabel array, and causes an exception that hangs the data grid display.

### DIFF
--- a/src/Oro/Bundle/DataGridBundle/Resources/public/js/datagrid/page-size.js
+++ b/src/Oro/Bundle/DataGridBundle/Resources/public/js/datagrid/page-size.js
@@ -110,7 +110,7 @@ define([
                 _.bind(
                     function(item) {
                         return item.size === undefined ?
-                            this.collection.state.pageSize === item : this.collection.state.pageSize === item.size;
+                            this.collection.state.pageSize == item : this.collection.state.pageSize == item.size;
                     },
                     this
                 )

### CODE REVIEW

**Comment by orocla** [comment]

File: src/Oro/Bundle/DataGridBundle/Resources/public/js/datagrid/page-size.js [hunk 1]
```diff
@@ -110,7 +110,7 @@ define([
                 _.bind(
                     function(item) {
                         return item.size === undefined ?
-                            this.collection.state.pageSize === item : this.collection.state.pageSize === item.size;
+                            this.collection.state.pageSize == item : this.collection.state.pageSize == item.size;
                     },
```
Grygir: This solution does not fit requirements of JS code style. Abstract Equality Comparison is restricted by adopted JSCS setup.

mbessolov: @Grygir So what kind of fix will be acceptable? Explicit string to number type conversion here and then strict equality checking?

cmenning:
@Grygir pointed out on the related issue #396 that this is not needed any longer due to a change in the handling of the filter results. I developed it against 1.8, and didn't notice the change in the master branch.

src/Oro/Bundle/DataGridBundle/Resources/public/js/datagrid/page-size.js
Grygir: @mbessolov, the issue is already fixed in 1.9.* and soon it will be in 1.8.3.

And, yes, for similar cases there have to be explicit type conversion.

PR #398, opened 2016-02-15
Update inline-actions.less

startupz:
If you don't add this line, box-shadow will also affect no-actions fields.


Grygir:
Hi @startupz,

Thank you for the contribution. The PR will be merged shortly.

PR #405, opened 2016-03-03
Extract oro/property-access

z38:
At the moment `oro/config` depends on classes from the property access component. In order to reference these classes, `oro/property-access` should be extracted into a subtree split and uploaded to Packagist.

I will submit a separate PR to incorporate these changes into the config component.


yurio:
Hi, @z38. Thank you very much for your pull request! It was merged to development branch, soon you'll see it at master.

PR #445, opened 2016-04-27
Remove hardcoded default to UTC timezone

### LINKED ISSUE #444: Twig DateTimeExtension is forced to UTC
Our system uses the HST timezone. One of our agents recently brought to my attention the fact that the MyEmails datagrid is reporting the email times in UTC and not HST.

I confirmed this on orocrm/platform 1.9.1 today by sending an email at 1:43PM HST, and here is what the MyEmails page shows:

screen shot 2016-04-26 at 6 19 41 pm

So I did some dumping from `/vendor/oro/platform/src/Oro/Bundle/EmailBundle/Resources/views/macros.html.twig` with this:

```PHP
{{ dump(date()) }}
{{ dump(date(date)) }}
{{ dump(date()|oro_format_date) }}
{{ dump(date(date)|oro_format_date) }}
```
I get this:
```PHP
// This is HST
object(DateTime)#21121 (3) {
  ["date"]=>
  string(26) "2016-04-26 15:47:58.000000"
  ["timezone_type"]=>
  int(3)
  ["timezone"]=>
  string(16) "Pacific/Honolulu"
}
// This is HST
object(DateTime)#21121 (3) {
  ["date"]=>
  string(26) "2016-04-26 13:43:08.000000"
  ["timezone_type"]=>
  int(3)
  ["timezone"]=>
  string(16) "Pacific/Honolulu"
}
// This is UTC
string(12) "Apr 27, 2016"
// This is UTC
string(12) "Apr 26, 2016"
```
The `date` parameter that is passed in to the twig macro is in UTC. The date() and date(date) calls correctly convert to the system timezone(HST). The problem is that when an HST date is passed to the oro_format_date filter it returns values in UTC and not HST. I don't believe this is the desired default behavior.

The problem appears to be in `/vendor/oro/platform/src/Oro/Bundle/LocaleBundle/Twig/DateTimeExtension.php`

There are 3 methods in that class that all contain this: `$timeZone = $this->getOption($options, 'timeZone', 'UTC');`

It appears that if you don't specify a timezone Option(which the twig macro does not) then it forces timezone to UTC, even if that is not what the system uses globally.

I solved this problem in my app by removing the 3rd parameter in all of the `getOption` calls.  I will submit a pull request, but I'm not sure if this will cause problems elsewhere.  I don't think so, because I can't think of a good reason you would always want to force to UTC if an option hasn't been passed.

codesman:
This was causing inaccurate date/time display in the MyEmails datagrid because the twig template is not set up by default to pass a timezone option and this code was set to force the timezone to UTC if there was no timezone option passed in.

Fixes #444

### DIFF
--- a/src/Oro/Bundle/LocaleBundle/Tests/Unit/Twig/DateTimeExtensionTest.php
+++ b/src/Oro/Bundle/LocaleBundle/Tests/Unit/Twig/DateTimeExtensionTest.php
@@ -102,7 +102,7 @@ class DateTimeExtensionTest extends \PHPUnit_Framework_TestCase
         ];
 
         $this->formatter->expects($this->once())->method('formatDate')
-            ->with($value, $dateType, $locale, $timeZone ?: 'UTC')
+            ->with($value, $dateType, $locale, $timeZone)
             ->will($this->returnValue($expected));
 
         $this->assertEquals($expected, $this->extension->formatDate($value, $options));
@@ -139,7 +139,7 @@ class DateTimeExtensionTest extends \PHPUnit_Framework_TestCase
         $options = ['locale' => $locale];
 
         $this->formatter->expects($this->once())->method('formatDay')
-            ->with($value, $dateType, $locale, $timeZone ?: 'UTC')
+            ->with($value, $dateType, $locale, $timeZone)
             ->will($this->returnValue($expected));
 
         $this->assertEquals($expected, $this->extension->formatDay($value, $options));
@@ -182,7 +182,7 @@ class DateTimeExtensionTest extends \PHPUnit_Framework_TestCase
         ];
 
         $this->formatter->expects($this->once())->method('formatTime')
-            ->with($value, $timeType, $locale, $timeZone ?: 'UTC')
+            ->with($value, $timeType, $locale, $timeZone)
             ->will($this->returnValue($expected));
 
         $this->assertEquals($expected, $this->extension->formatTime($value, $options));
--- a/src/Oro/Bundle/LocaleBundle/Twig/DateTimeExtension.php
+++ b/src/Oro/Bundle/LocaleBundle/Twig/DateTimeExtension.php
@@ -91,7 +91,7 @@ class DateTimeExtension extends \Twig_Extension
     {
         $dateType = $this->getOption($options, 'dateType');
         $locale = $this->getOption($options, 'locale');
-        $timeZone = $this->getOption($options, 'timeZone', 'UTC');
+        $timeZone = $this->getOption($options, 'timeZone');
 
         return $this->formatter->formatDate($date, $dateType, $locale, $timeZone);
     }
@@ -113,7 +113,7 @@ class DateTimeExtension extends \Twig_Extension
     {
         $dateType = $this->getOption($options, 'dateType');
         $locale = $this->getOption($options, 'locale');
-        $timeZone = $this->getOption($options, 'timeZone', 'UTC');
+        $timeZone = $this->getOption($options, 'timeZone');
 
         return $this->formatter->formatDay($date, $dateType, $locale, $timeZone);
     }
@@ -136,7 +136,7 @@ class DateTimeExtension extends \Twig_Extension
     {
         $timeType = $this->getOption($options, 'timeType');
         $locale = $this->getOption($options, 'locale');
-        $timeZone = $this->getOption($options, 'timeZone', 'UTC');
+        $timeZone = $this->getOption($options, 'timeZone');
 
         return $this->formatter->formatTime($date, $timeType, $locale, $timeZone);
     }


DimaSoroka:
Hello @codesman, there [4 failed tests](https://travis-ci.org/orocrm/platform/jobs/126145272) related to this PR, could you please take a look?

codesman:
Hello @DimaSoroka,

Doh!  Should have known better than to push without testing!

vsoroka:
Hi @codesman ,

It seems that the current implementation is valid because Date and Time data types are timezone agnostic (e.g. November 15 is always November 15 and 11:00 PM is always 11:00
PM no matter what timezone is set on a server or on System > Configuration > Localization page of your site). DateTime data type has timezone and we displays it with considering timezone is set on System > Configuration > Localization page. The date displayed on My Emails page are stored as DateTime, we just display it in different formats (time, "Yesterday" or date), it depends on a time when an email was sent.

I've tried to reproduce the problem described in https://github.com/orocrm/platform/issues/444, but I didn't.
Based on the provided description I can say that:
- The server timezone is HST
- The email was sent at 1:43PM HST, but it is not clear the date
- It is not clear which timezone is set on System > Configuration > Localization page

Could you please provide us steps to reproduce the issue?

codesman:
Hello @vsoroka,
## steps to reproduce:
   * git clone https://github.com/orocrm/platform-application.git oro_bap
   * cd oro_bap/
   * git checkout -b 1.9.2
   * composer install --prefer-dist --no-dev
   * accept all parameter defaults
   * php app/console doctrine:database:create
   * php app/console oro:install --env prod
   * cd web/
   * php ../app/console server:run
   * login admin user at http://127.0.0.1:8000/user/login
   * go to config/system/platform/localization and confirm that HST is the System Timezone
    screen shot 2016-05-04 at 8 54 21 am  
   * edit admin user email synchronization settings and enter credentials for an existing email account
   * confirm that the connection to the email server is successful
   * send an email to the admin user indicating current System/HST time
   * go to email/user-emails and "Sync Emails"
   * go to dashboard/view/1 and see this:
screen shot 2016-05-04 at 8 57 52 am  
   * Yay! The displayed time is correct!  Now go to email/user-emails and see this:
screen shot 2016-05-04 at 8 58 44 am  
   * Oh noes! This is wrong!  It is displaying the time 10 hours later.

I understand that all times are stored agnostically in UTC and(are supposed to be) displayed based on System Localization settings.  The problem appears to be that the filter that is used for display is not allowing System Timezone to be displayed by default. If the system localization is set to use GMT/UTC Timezone, then I assume the current implementation would appear to work perfectly!

The `DateTimeExtension->formatDate()` method that `oro_format_date` filter uses for the display in the MyEmails datagrid is implemented in a way that if a Timezone option is not passed in to the filter, then it "decides" that you want it to return 'UTC' and not the System Timezone.  This is not true!  We always want it to return values in the System Timezone unless an option is passed in for another Timezone, right?

If the calls to `oro_format_date` in the Twig macro `date_smart_format(date)` do not implicitly pass in a Timezone(System or other) option then the value is forced to be returned/displayed in UTC. The only solution to fix this would be to edit/override the  `date_smart_format(date)` macro in  `vendor/oro/platform/src/Oro/Bundle/EmailBundle/Resources/views/macros.html.twig` and pass in whatever Timezone the System is using on every call to `oro_format_date`.  Surely, requiring everyone that is not using UTC to hardcode their System Timezone into this macro is not the answer!

Also notice that the `oro_format_datetime` filter properly displays the System Timezone by default, because the `DateTimeExtension->formatDateTime()` method does not default to 'UTC' if no Timezone option is passed in, as is not the case in the other 3 methods: `DateTimeExtension->formatDate(), DateTimeExtension->formatDay() and DateTimeExtension->formatTime()`.  

Hope that helps!
Tom

vsoroka:
Thank you for detailed STR, I will check it ASAP.

vsoroka:
Hi @codesman ,

It seems that the issue is more complex than hardcoded UTC timezone in DateTimeExtension. This hardcode looks correct.

After investigation I've found the following:
- The My Emails and Dashboard display two different dates. My Emails displays "sent" date, but Dashboard displays "received" date. It is one part of the issue.
- The "sent" and "received" dates can be very different. It seems in your case the difference is 10 hours. It is the main reason of the issue. But, I guess, it will be not easy to find a cause of it.

As I still cannot reproduce the issue on my local environment I need your help.
Could you please provide the following information:
- a value of "sent" column from "oro_email" table and "received" column from "oro_email_user" table for the test email.
- a dump of headers of the test email. You can get this info in your email client. Here is an example of such info (actually I need all "Received" headers and "Date" header):

```
From ...email address... Thu May  5 08:53:19 2016
Received: from [...] (helo=...) by ...;
 Thu, 05 May 2016 08:53:19 +0300
Received: from [...] (helo=...) by ...
 QID:1ayCDv-0008rF-LT/RC:1; Thu, 05 May 2016 08:53:19 +0300
Received: from [...] (helo=[...]) by ...
 with esmtpsa ID 1ayCDv-000NKZ-I7 for ...email address...; Thu, 05 May 2016
 08:53:19 +0300
Message-ID: <...>
Subject: test3
From: Vova Soroka <...email address...>
To: ...email address...
Content-Type: text/plain
Date: Thu, 05 May 2016 08:43:42 +0300
Mime-Version: 1.0
X-Mailer: Evolution 3.10.4-0ubuntu2 
X-Evolution-Source: 1462425901.3236.2@vsoroka
Content-Transfer-Encoding: 8bit
```

codesman:
Hi @vsoroka,

```SQL
SELECT * FROM oro_email_user WHERE id=1;
```
```JSON
[
  {
    "id": 1,
    "email_id": 1,
    "mailbox_owner_id": "NULL",
    "created_at": "2016-05-04 15:57:07",
    "received": "2016-05-04 15:56:34",
    "is_seen": 0,
    "organization_id": 1,
    "user_owner_id": 1,
    "origin_id": 1
  }
]
```


```SQL
SELECT * FROM oro_email WHERE id=1;
```
```JSON
[
  {
    "id": 1,
    "from_email_address_id": 3,
    "thread_id": "NULL",
    "email_body_id": 1,
    "created": "2016-05-04 15:57:07",
    "subject": "Testing the time display - 5:56AM HST",
    "from_name": "Tom Holland <sender@example.com>",
    "sent": "2016-05-04 15:56:32",
    "importance": 0,
    "internaldate": "2016-05-04 15:56:34",
    "message_id": "<B5EA0BEA-450D-4BC9-AF68-03BB50A83C02@example.com>",
    "x_message_id": 0,
    "x_thread_id": 0,
    "is_head": 1,
    "refs": "NULL",
    "multi_message_id": "NULL",
    "acceptLanguageHeader": 0
  }
]
```
> The My Emails and Dashboard display two different dates. My Emails displays "sent" date, but Dashboard displays "received" date. It is one part of the issue.
> The "sent" and "received" dates can be very different. It seems in your case the difference is 10 hours. It is the main reason of the issue. But, I guess, it will be not easy to find a cause of it.

The values that are stored in the database for this email are:
   * `"oro_user_email.created_at": "2016-05-04 15:57:07"` // The time that email was synchronized from the recipient's server and the entry was inserted into the `oro_user_email` table  
   * `"oro_user_email.received: 2016-05-04 15:56:34"` // The time that the recipient's email server received the email
   * `"oro_email.created: 2016-05-04 15:57:07"` // Identical to `oro_user_email.created_at`, Is this a redundant column?
   * `"oro_email.sent: 2016-05-04 15:56:32"` // The time that the sender's email client sent the email   
   * `"oro_email.internaldate: 2016-05-04 15:56:34"` // Identical to `oro_user_email.received`, Is this a redundant column?

All of these values are within a timespan of 35 seconds, there is no difference of 10 hours anywhere. Except when it is displayed in the My Emails list.

Here are the headers from the email:
```
Delivered-To:	recipient@example.com
Return-Path:	<sender@example.com>
Received:	from email.server by email.server (Dovecot) with LMTP id hNXTMvwHKlcoQAAA7aV/mQ for <recipient@example.com>; Wed, 04 May 2016 11:56:34 -0400
Received:	from email.server (using TLSv1 with cipher DHE-RSA-AES256-SHA (256/256 bits)) by email.server (Dovecot) with LMTP id DI68AlqUDVeaMwAAe3+vfg ; Wed, 04 May 2016 11:56:34 -0400
Received:	from email.server by email.server (envelope-from <sender@example.com>) (ecelerity 2.2.3.49 r(42060/42061)) with ESMTPS (cipher=AES256-SHA) id 83/F0-23528-1BB1A275; Wed, 04 May 2016 11:56:33 -0400
Received:	from email.server (localhost.localdomain [127.0.0.1]) by email.server (SMTP Server) with ESMTP id BBF55380580 for <recipient@example.com>; Wed, 4 May 2016 11:56:33 -0400 (EDT)
X-Auth-ID:	sender@example.com
Received:	by email.server (Authenticated sender: sender-AT-example.com) with ESMTPSA id 890C63804E1 for <recipient@example.com>; Wed, 4 May 2016 11:56:33 -0400 (EDT)
X-Sender-Id:	sender@example.com
Received:	from sender.ip (using TLSv1 with cipher DHE-RSA-AES256-SHA) by 0.0.0.0:587 (trex/5.5.4); Wed, 04 May 2016 11:56:33 -0400
From:	Tom Holland <sender@example.com>
Content-Type:	text/plain; charset=utf-8
Content-Transfer-Encoding:	quoted-printable
Subject:	Testing the time display - 5:56AM HST
Message-Id:	<B5EA0BEA-450D-4BC9-AF68-03BB50A83C02@example.com>
Date:	Wed, 4 May 2016 08:56:32 -0700
To:	Tom Holland <recipient@example.com>
```

Couple things to note:
   * The sending and receiving email are 2 different addresses, not the same one.
   * The test email was sent from an external, desktop email client, not from the test install
   * The email server for both addresses is a large cloud provider and they are on EDT(UTC -0400)
   * The test email was sent from the PDT(UTC -0700) timezone, not HST.
   * The difference between PDT and HST is -0300

I certainly could be barking up the wrong tree, but it still appears to me that the problem is that the Twig Filter is not capable of returning non-UTC values unless explicitly told to.  So, we have this code...

```PHP
public function formatDate($date, array $options = [])
{
    $dateType = $this->getOption($options, 'dateType');
    $locale = $this->getOption($options, 'locale');
    $timeZone = $this->getOption($options, 'timeZone', 'UTC');

    return $this->formatter->formatDate($date, $dateType, $locale, $timeZone);
}

protected function getOption(array $options, $name, $default = null)
{
    return isset($options[$name]) ? $options[$name] : $default;
}
```

Given the way this code is designed, it is **impossible** to get back anything but UTC values from `{{ date(date) | oro_format_date }}`. Unless you call `{{ date(date) | oro_format_date( ['timeZone' => 'someTimezone'] ) }}`, `date(date)` will be returned in UTC. 'UTC'  **always** gets passed to the formatter in the absence of an option. This seems wrong.

Can you explain why this implementation looks correct to you?  I cannot understand why this would be the desired behavior. What am I missing?

Tom

vsoroka:
Hi @codesman ,

About the issue: it seems that you use the platform v1.9.1 (see https://github.com/orocrm/platform/blob/1.9.1/src/Oro/Bundle/EmailBundle/Resources/views/macros.html.twig#L488-L500). This issue was fixed in v1.9.2 (see https://github.com/orocrm/platform/blob/1.9.2/src/Oro/Bundle/EmailBundle/Resources/views/macros.html.twig#L488-L500)

About hardcode UTC timezone: we did this because it was supposed that `oro_format_date` and `oro_format_day` TWIG filters will be used with Date data type, and `oro_format_time` with Time data type, which is timezone agnostic. For example it is used for user's birthday, opportunity's close date, etc. These values are stored as Date, not as DateTime, and they should be displayed as is, without conversion to system timezone. As PHP does not have separate classes for Date and Time data types, we cannot determine which data type is used to store of such values. What we can deal with this:
- current approach: if you want to format Date and Time values you just use `value|oro_format_date` or `value|oro_format_time`. If you want to format the date part of DateTime value you have to specify timezone directly, e.g. `value|oro_format_date({'timeZone': oro_timezone()})`.
- proposed approach: if you want to format Date and Time values you have to specify UTC timezone directly, e.g. `value|oro_format_date({'timeZone': 'UTC'})`, but to format the date part of DateTime value these filters can be used without timezone, e.g. `value|oro_format_date`.

I think both approaches are not obvious and it is hard to say which is better. The main disadvantage of the proposed approach is a backward compatibility break. What do you think about it?

codesman:
Hi @vsoroka ,

> About the issue: it seems that you use the platform v1.9.1

Yes, the system that the problem was reported on is using 1.9.1. I have now upgraded my local dev environment to 1.9.3 and confirmed that the issue is fixed, will push 1.9.3 to production.  Time to shut this PR down. Thanks for working through this with me!

I thought I had reproduced the issue on a clean install of 1.9.2, but apparently not...

> About hardcode UTC timezone:

Thank you again for the detailed explanation. I understand the problem and why the decisions were made.  Perhaps a little something should be added to the docblock to clarify?

> I think both approaches are not obvious and it is hard to say which is better.

Agree! Given the constraints, the current approach seems as valid as the proposed approach and since there isn't a clear and compelling advantage to the proposed approach, I'd say it's not worth making a breaking change for it.

PR #454, opened 2016-05-10
Permit the value "Zero" as key in an enum array

Djamy:
Currently, if we put a zero as a key in enums arrays, the key is set as equal to the value. 
But, we might need to have a zero as a key/id.

So, by changing the empty() by !isset, we can set the value "0" as an id.

I also included a unit test.


zebimax:
Hi, @Djamy. Thank you for your contribution. But changing the empty() by !isset, we will also allow to create enum values with empty strings as $id and this will cause an errorы in processing such enums.

zebimax:
Thanks again for your contribution. Soon you'll see it at master.

iklymenko:
@Djamy, your changes merged. Thanks for your contribution!

PR #465, opened 2016-06-01
Update wamp.js

sadortun:
Do not close the session if `this.session` is null


Grygir:
Hi @sadortun,

Thank you for the contribution. Changes are merged in to internal repo and soon they will be in public.

Issue #476, opened 2016-06-24
CSV Export Bug (Excel) with UTF-8 BOM and multiline Fields


sbani:
Hi all,

exporting a datagrid to a csv (from a datagrid.yml grid or a report) leads to multiple errors in excel.
Excel is not very tolerant with csv, but nearly everyone who wants to export a csv opens it in excel.

Bug 1:
BOM is missing. That's why you see excel encoding errors.

Bug 2:
The newlines within a field are made with `\n\r`. This is not correct and should be `\n` only.

I start working no this **now** and will create a PR soon.

sbani:
Further read: http://stackoverflow.com/questions/15974163/export-csv-for-excel
Excel makes me going crazy. 

Do you have any further knowledge here? Perhaps the separator should be `\t` too?

artml:
@sbani thanks for helping us on this. I guess the proper solution to this problem will be an Excel-specific export, but that's another task for the future.

What Excel version do you use for import by the way?

sbani:
We need:
Excel for Mac v15
Excel for Windows: Excel 2006 and Excel 2003

Im added an excel specific export! It works great but I have problems when batchSize is exceeded because I can't just echo to `php://output`. I think I need to create a tmp file and send to php://output afterwards

artml:
@sbani wow, that's some cool addition!

Do you want to submit the PR, or create an extension for the marketplace?

sbani:
sure. I will post my PR asap.
But I need to create a better code base for more options and I need better tests. CSV and XLS(X) should probably have the same base classes. At the moment it's just a "quick fix"

sbani:
I added the export. Hope it's enough. I created everything for `1.9`, so haven't tested in UI for `1.10`.
This should fix most of the "excel" problems but the csv bugs are still present. 

As I stated in the PR: A refactor should be done, because it makes sense to use the same code base for all kind of csv, xls, xlsx file creation.

I could do it but I first have to implement the hole thing for our AppBundle because we need it now and I can't wait for 1.10 (plus the implementation)

artml:
@sbani thanks! We will check it for 1.10 branch as well.

PR #477, opened 2016-06-24
Add UTF-8 BOM to csv output.

sbani:
Adding the BOM is mandatory if you use utf-8 and want to display the csv with Microsoft/Excel.

Related to #476


sbani:
I will close here because I think the csv export works fine and it's better to add an additional "excel export" controller/option

sergeyz:
Hi, @sbani!

You are right. All you need is writer + batch job, pass batch job as a button configuration - should work fine.

Thanks for update!

PR #481, opened 2016-06-27
Added xlsx (excel) datagrid export

### LINKED ISSUE #476: CSV Export Bug (Excel) with UTF-8 BOM and multiline Fields

sbani:
It is now possible to export a datagrid (batch or not) as an xlsx (Excel 2007) file.
- Added tests
- Created base classes for the overlap between csv and xlsx.
- Added yaml configurations
- Added the `liuggio/ExcelBundle` for the excel service which is based on `phpoffice/phpexcel`
- Added the Bundle to the `OroImportExportBundle` bundles.yml because it's basically used there.

## TODO
- Create option to export "xls" instead of "xlsx".
- Create CSV with the same excel classes(?)

Closes #476


mkudelya:
Hi @sbani ,
We have reviewed your pull request and have tested it. Export is working, but orocrm should work on 512MB memory limit. I created 10000 accounts and ran export, after 45 seconds i got fatal error:

Fatal error:  Allowed memory size of 536870912 bytes exhausted (tried to allocate 85 bytes) in /Library/WebServe
r/www/symfony/orocrmee_source/vendor/phpoffice/phpexcel/Classes/PHPExcel/CachedObjectStorage/CacheBase.php</b> on line 186

Please fix this issue

sbani:
Thx @mkudelya for your tests. Found the bug (memory leak).

I basically removed the file write to the tmp file because it's not needed. I enrich the excel object step by step and write on `close()` to the file.

Though you have to know that the memory usage is always higher than writing CSV because there is no way to append data to the file (because it's zipped, etc).

But I tried it with around 12k accounts and it worked fine.

mkudelya:
@sbani 
I can offer one decision, if grid contains more than 10k rows, before export orocrm will be show popup where you can choose, either cancel operation or continue at one's own risk. Similar popup exists on mass delete action.

sbani:
sounds good. do I need to implement it? I prefer not to write frontend code if that's possible

mkudelya:
@sbani we will do frontend on our side. thanks!

sbani:
thank you

artml:
Closing the PR as we've added missing functionality and merged it to master. @sbani thanks again for your contribution!

PR #487, opened 2016-07-05
Be more verbose about the error during a Delete

### LINKED ISSUE #486: [Bug] 500  response during DELETE display "Permision Denied"
When a Ajax DELETE request return a `500`, 

image

The error handler assumes that the permision was denied.
image

sadortun:
Fixes #486 

When a Ajax DELETE request return a `500`, 

image

The error handler assumes that the permision was denied.
image

This PR changes the default message to display the message provided by the API

image


mkudelya:
Fix was merged! Thank!

PR #489, opened 2016-07-05
Add missing (but used) object property

sbani:
The property `resource` is missing in `FolderContentCumulativeLoader`.
Added the property because it is used within the object


mkudelya:
@sbani pull request is merged. Thanks!

PR #495, opened 2016-07-08
Prevent accessing null object

sadortun:


mkudelya:
@sadortun, changes are merged, thanks for your contribution!

PR #147, opened 2014-09-05
Extracted OroRequireJS bundle as standalone bundle

ste93cry:
This should fix #100. Please review my changes before merging

PR #496, opened 2016-07-09
Make RequireJSBundle independent from CacheBundle

ste93cry:
This PR continues the work started in #147


ste93cry:
@sergeyz could you review this PR instead of the old one? It should be fine, tests fails but seems to not be related to my edits

src/Oro/Bundle/RequireJSBundle/DependencyInjection/OroRequireJSExtension.php
```diff
@@ -29,5 +30,15 @@ public function load(array $configs, ContainerBuilder $container)
...
         $container->setParameter('oro_require_js.build_path', $config['build_path']);
+
+        if (!$container->hasDefinition('oro.cache.abstract') && $container->hasDefinition('oro_requirejs.cache')) {
+            $oldDefinition = $container->getDefinition('oro_requirejs.cache');
+
+            $definition = new Definition('Doctrine\\Common\\Cache\\PhpFileCache', array('%kernel.cache_dir%/oro_data'));
+            $definition->setMethodCalls($oldDefinition->getMethodCalls());
+            $definition->setPublic(false);
+
+            $container->setDefinition('oro_requirejs.cache', $definition);
```
sergeyz: Lets use decoration here http://symfony.com/doc/current/service_container/service_decoration.html

ste93cry: PR updated with the changes requested

sergeyz: Thank you!

src/Oro/Bundle/RequireJSBundle/DependencyInjection/Compiler/CacheProviderPass.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+class CacheProviderPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->hasDefinition('oro.cache.abstract')) {
+            $definition = new Definition('Doctrine\\Common\\Cache\\PhpFileCache', array('%kernel.cache_dir%/oro_data'));
```
sergeyz: Feel free to use PhpFileCache::class here

ste93cry: I would like to do it, but Symfony 2.7 (which is declared to be the minimum version supported for this bundle in `composer.json`) runs on PHP 5.3+ while the `::class` syntax has been available since PHP 5.5

ste93cry:
@sergeyz When do you think this can be merged? I would like to use this bundle for a project I'm working on

sergeyz:
@ste93cry working on merge

ste93cry:
By having the constraint `2.8.*` for the Symfony entry, who has a greater or lower version won't be allowed to use this bundle, right? As it seems that it does not really depend on any other bundle, maybe you could change that constraint to something more relaxed, at least `~2.7`

sergeyz:
Sure, php 5.5 is ok for you?

ste93cry:
Yes it's fine. I'm using anyway Symfony 2.8, so the minimum version of Symfony that I proposed was just an example

sergeyz:
HI, @ste93cry !

After discussion with @vsoroka I've tried to avoid compiler pass and added services support to app.yml. Fix already in master, but, unfortunately, not in 1.10. It will be available in 1.10 branch shortly and in 1.10.1 release

ste93cry:
Hi, I'm glad to see that you managed to find a better way to handle this PR. Please remember to sync the standalone repository too when it's time. Although this isn't the right place where to discuss this, I would like to report that there is still that I missed to see before: the PhpUtils component.

sergeyz:
I'm going to work on split automation in next few days, lets keep issue open until I do it

ste93cry:
@SergeyCherepanov any news on this? I would like to be sure this does not get missed, as merging the PR "closed" it (it does not show up anymore in the list)

sergeyz:
Hi, @ste93cry! It not missed, its in todo list

sergeyz:
@ste93cry today I pushed actual code to https://github.com/orocrm/RequireJSBundle. Had to switch branches because of sync issues. It looks like there will be issue with assets, we use https://github.com/fxpio/composer-asset-plugin to load them but composer.json is outdated.

Grygir:
Unfortunately, README.md misses the part that `composer.json` of an application have contain this configuration
```json
  "extra": {
    "asset-installer-paths": {
      "npm-asset-library": "web/bundles/npmassets"
    }
  }
```
That defies the place for npm assets in public directory.
We have to update installation guide for RequireJS bundle

ste93cry:
> we use https://github.com/fxpio/composer-asset-plugin

Although for your use-case it may be better to use such a plugin, usually other tools are used to manage frontend dependencies (bower, npm, yarn). What about having a configuration option to set where the `r.js optimizer` is located? This would give users flexibility over which tool to use to install such assets

Grygir:
You are right, path to library location have to be configurable. We will add this improvement to our back log. But I'm not sure how soon when it'll be done. So if you are up for this, the contribution is highly appreciated.

ste93cry:
Ok, I will try to see if I find the time to contribute the change. Another thing is that (even if it's not in the `composer.json`) the bundle requires the `PhpUtils` component, which is not yet extracted as a standalone repository

PR #502, opened 2016-07-15
Upgraded Doctrine ORM to 2.5.4.

mathop:
The current Doctrine ORM version (2.5.1) doesn't work with PHP 7 type hints, it was fixed in the version 2.5.3.

[DDC-4019: Proxy generator is not including PHP7 return type hints](https://github.com/doctrine/doctrine2/issues/4884)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "ext-xml": "*",
         "symfony/symfony": "2.7.*",
         "twig/twig": "1.20.0",
-        "doctrine/orm": "2.5.1",
+        "doctrine/orm": "2.5.4",
         "doctrine/doctrine-bundle": "~1.6.0",
         "doctrine/dbal": "2.5.1",
         "doctrine/data-fixtures": "1.1.1",


orocla:

mkudelya:
2.0 version of platform is already contains 2.5.5 version of doctrine/orm.
Thank you!

anyt:
1.10 compatible with PHP 7.0 so issue above should not be a problem there.

As I know It is not planned yet to upgrade Doctrine ORM for 1.10. But even if it will be implemented you should know that for Community Edition we provides patch updates only for the latest LTS version, currently, this is 2.6 LTS. So even if an update will be scheduled it will be done only on Enterprise version.

dewos:
1.10 is compatible with PHP 7.0, but doctrine/orm 2.5.1 is not.

anyt:
yes, it doesn't support new features, but if you'll not use type-hint with php 7.0 this will work.

Issue #520, opened 2016-08-30
An exception in createSynchronizationProcessor causes imap-sync to skip all other email origins


bluedot74:
In the doSyncOrigin method in AbstractEmailSynchronizer the createSynchronizationProcessor method is called. If an exception occurs inside this method the EmailOrigin's sync code is not set to "SYNC_CODE_FAILURE".
Since the sync code is not set to failure the findOriginToSync grabs the same EmailOrigin it just attempted. 
This causes the origin to fail the check:

```
          if (isset($processedOrigins[$origin->getId()])) {
                $this->logger->info('Exit because all origins have been synchronised.');
                break;
            }
```

Which breaks the loop, and all the other email origins are skipped.

bluedot74:
I submitted a PR to try to fix this: https://github.com/orocrm/platform/pull/521

artml:
@bluedot74 thank you for your contribution! We will review this PR shortly

PR #521, opened 2016-08-30
Fix for Issue 520 - Exception during createSynchronizationProcessor causes imap-sync to skip all other email origins

bluedot74:
An attempt to fix: https://github.com/orocrm/platform/issues/520


mkudelya:
Bug already has been fixed.

PR #548, opened 2016-10-11
OroEntityExtendBundle::checkConfig() Process error sent to stdout

sadortun:
Hello Oro Team,

When an exception occur during `oro:entity-extend:cache:check` the message is sent to `stdout` and not `stderr`

image

image

Have a good day,
Samuel

### DIFF
--- a/src/Oro/Bundle/EntityExtendBundle/OroEntityExtendBundle.php
+++ b/src/Oro/Bundle/EntityExtendBundle/OroEntityExtendBundle.php
@@ -122,7 +122,10 @@ class OroEntityExtendBundle extends Bundle
                 $process = $pb->getProcess();
                 $exitStatusCode = $process->run();
                 if ($exitStatusCode) {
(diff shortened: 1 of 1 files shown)


mkudelya:
Hi @sadortun 

For example when 'data' column in oro_entity_config is corrupted (and also with other fatals) i get errors from getErrorOutput()

sadortun:
Hello @mkudelya , Ok,  but since not all errors go to Stderr I think this product is still valid.

mkudelya:
What do you think about this code? I think we cover two cases.

if ($exitStatusCode) {
$output = $process->getErrorOutput();

if (empty($output) {
$output = $process->getOutput();
}

throw new \RuntimeException($output);
}

sadortun:
@mkudelya  It was supposed to be there, (check the first screen capture on the original PR)

Note to self: Dont do merge PR on a tablet using github UI while on the road !

mkudelya:
@sadortun 

This PR has been merged. Thank you.

PR #553, opened 2016-10-21
Added direct @service qb support to datagrids configuration

dewos:
Not everyone likes to mess-up with query in YAML or the Doctrine getRepository(Entity::class) locator pattern. 

This pull request allows much more flexibility in datagrids configuration.

``` yml
datagrid:
    users-grid:
        source:
            type: orm
            service: '@acme_bundle.users.custom.repository->getUsersQueryBuilder()'

```

### DIFF
--- a/src/Oro/Bundle/DataGridBundle/Datasource/Orm/Configs/YamlProcessor.php
+++ b/src/Oro/Bundle/DataGridBundle/Datasource/Orm/Configs/YamlProcessor.php
@@ -37,7 +37,8 @@ class YamlProcessor implements ConfigProcessorInterface
             } else {
                 throw new DatasourceException(
                     sprintf(
-                        'Datagrid configured with service must return an instance of Doctrine\ORM\QueryBuilder, %s given',
+                        '%s configured with service must return an instance of Doctrine\ORM\QueryBuilder, %s given',
+                        get_class($this),
                         is_object($qb) ? get_class($qb) : gettype($qb)
                     )
                 );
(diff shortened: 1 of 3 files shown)


orocla:

mkudelya:
Hi @dewos 

Thank you for your contribution! I have small note about our code. I think will be better if you change 'service' to 'queryBuilder'. What do you think?

mkudelya:
@dewos 

Another thing, you can get qb only for 'orm' type.

dewos:
Hi @mkudelya,  

1) Are you sure about 'queryBuilder'? Yaml normally follows snake-case, not the camelCase (hence 'query-builder'). Moreover it seems to me that 'source' (where?) -> 'query'|'service' is more descriptive. 

2)  Yes, but the YamlProcessor is only for 'orm' right now, I'm right?  Should I also globally check it?

mkudelya:
Hi @dewos 

1. As i see in datagrid.yml we are using both variants of case. 'service' option is good when you have one source, but if we will have more, for example in future we will add other source, option 'service' will be unclear. In other case, is source of what ?

2. Sorry, you are right.

3. Don't forget change the documentation, for example [here](https://github.com/orocrm/platform/blob/master/src/Oro/Bundle/DataGridBundle/Resources/doc/backend/datagrid.md)

dewos:
@mkudelya 

1) Wait... All _Options_ in datagrid.yml are snake_case (**data_name**, **frontend_type**, **bind_parameters** etc...). Data _columns_ can be camelCased (**firstName**, lastName etc..), but they are getter oriented, not options. We should keep this convention.

1b)  I am not sure that a single datagrid might works with two different "sources" **at the same time**. Anyway  the **query_builder** key is good to me, if you like it too.

2) Good

3) I hope to find time :)

mkudelya:
1) 2) Ok :)

Thanks you!

dewos:
@mkudelya here we go. Sorry for the delay... tough week. 
 Should we backport this one also in the 1.10 branch? What is the ETA for 2.0, any way?

P.s.
No idea why the build fails, but is not a problem on my side, I guess.

mkudelya:
Hi

I have already merged this feature. It will be available in the next release.
We plan 2.0 release at the end of the year.
Thanks!

PR #573, opened 2016-11-24
Fix #547 Add migration to use InnoDB on mysql version 5.6.

### LINKED ISSUE #547: OroSearch use MySAM schema that is incompatible with MySQL GTID / Replication
Hello Oro team,

While deploying on a Mysql 5.7 server that have replication configured, it look like the install script fail.

```
[info] >> process triggers modifications stored in DB
Loading "main" data fixtures ...
  > loading [-240] Oro\Bundle\OrganizationBundle\Migrations\Data\ORM\LoadOrganizationAndBusinessUnitData

  [Doctrine\DBAL\Exception\DriverException]                                                                                                                                                                                                                                
  An exception occurred while executing 'INSERT INTO oro_search_index_text (field, value, item_id) VALUES (?, ?, ?)' with params ["name", "default", 5]:                                                                                                                   
  SQLSTATE[HY000]: General error: 1785 Statement violates GTID consistency: Updates to non-transactional tables can only be done in either autocommitted statements or single-statement transactions, and never in the same statement as updates to transactional tables.                                                  
```

current mysql options :
![image](https://cloud.githubusercontent.com/assets/794449/19257332/ac1ea130-8f3c-11e6-96b5-301b203b0a78.png)

Have a good day,
Samuel

sadortun:
OroSearch use MySAM schema that is incompatible with MySQL GTID / Replication

### DIFF
--- a/src/Oro/Bundle/SearchBundle/Migrations/Schema/v1_2/OroSearchBundle.php
+++ b/src/Oro/Bundle/SearchBundle/Migrations/Schema/v1_2/OroSearchBundle.php
@@ -4,48 +4,16 @@ namespace Oro\Bundle\SearchBundle\Migrations\Schema\v1_2;
 
 use Doctrine\DBAL\Schema\Schema;
 
-use Doctrine\ORM\EntityManager;
 use Oro\Bundle\MigrationBundle\Migration\Migration;
 use Oro\Bundle\MigrationBundle\Migration\QueryBag;
-use Symfony\Component\DependencyInjection\ContainerAwareInterface;
-use Symfony\Component\DependencyInjection\ContainerInterface;
 
-class OroSearchBundle implements Migration , ContainerAwareInterface
+class OroSearchBundle implements Migration
 {
-    /**
-     * @var ContainerInterface
-     */
-    protected $container;
-
-
-    /**
-     * @inheritdoc
-     */
-    public function setContainer(ContainerInterface $container = null)
-    {
-        $this->container = $container;
-    }
-
     /**
      * {@inheritdoc}
      */
     public function up(Schema $schema, QueryBag $queries)
     {
-        //save old avatars to new place
-        /** @var EntityManager $em */
-        $em         = $this->container->get('doctrine.orm.entity_manager');
-
-        $driverName = $em->getConnection()->getDriver()->getName() ;
-
-        if(in_array($driverName , array( 'pdo_mysql' ,'mysqli' ))) {
-
-            $query      = "show variables like \"version\"";
-            $version = $em->getConnection()->executeQuery($query)->fetchColumn(1);
-
-            if(version_compare($version , '5.6.0' , '>=')) {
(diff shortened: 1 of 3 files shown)


sergeyz:
Hi, @sadortun, thank you for the contribution!

One more check required here https://github.com/orocrm/platform/blob/master/src/Oro/Bundle/SearchBundle/EventListener/ORM/FulltextIndexListener.php#L49

src/Oro/Bundle/SearchBundle/Migrations/Schema/v1_2/OroSearchBundle.php
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function up(Schema $schema, QueryBag $queries)
+    {
+        //save old avatars to new place
+        /** @var EntityManager $em */
+        $em         = $this->container->get('doctrine.orm.entity_manager');
```
vsoroka: please do not use EntityManager in a migration. To get a connection the migration can implement `Oro\Bundle\MigrationBundle\Migration\ConnectionAwareInterface` interface

sadortun: No problem,  example was taken from https://github.com/orocrm/platform/blob/master/src/Oro/Bundle/UserBundle/Migrations/Schema/v1_2/OroUserBundle.php#L53

sadortun:
Hello @sergeyz  i did notice that it was use there, but `FulltextIndexListener` does not seems to be used by anybody, so my guess is that it's loaded automactly.

Your are probably the best fit to update this one.

Thanks

sadortun:
@vsoroka Updated Thanks for the info!

@sergeyz Yes i've noticed, i have very little time available next week(s), I am not sure why the class metadata is actually being overwriten, as you are more familiar with this i'll let you update the `FulltextIndexListener` and test for the expected behaviour.

It should look to something like :
```php
        $driverName = $this->connection->getDriver()->getName() ;

        $engine = PdoMysql::ENGINE_MYISAM;
        if(in_array($driverName , array( 'pdo_mysql' ,'mysqli' ))) {

            $version = $this->connection->executeQuery('show variables like "version"')->fetchColumn(1);

            if(version_compare($version , '5.6.0' , '>=')) {
                $engine = PdoMysql::ENGINE_INNODB;
            }
        }

        $classMetadata->table['options']['engine'] = $engine;
```

sergeyz:
@sadortun Looks good. One more thing we can improve is using `select version();` instead of `show variables like "version"`

One more thing I've noticed - we have a lot of manual search table cleanups because of non-transactional MyISAM. I'm not sure what should we do in this case. Looks like we have to leave it as is. 

I would say that the best option will be to set engine for this table using some configuration. And you will be able to change it for your application.

mkudelya:
Pull request has been merged. Thank you @sadortun

PR #579, opened 2016-12-06
Fixes #424 Tag Groups and taonomies

### LINKED ISSUE #424: Add Tag Groups as part of tags bundle
The ability to group tags together as Taxonomies/Tag Groups would allow admins to create more structured understanding of tags and sibling relationships to other tags.

sadortun:
As promised ! 

Rebased on 1.10.

Fixes #424

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ composer.lock
 /vendor
 phpunit.xml
 *~
+*.bak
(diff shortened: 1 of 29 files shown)


mkudelya:
This is only 2.0 version feature, we don't add new features to old releases.

PR #585, opened 2016-12-16
Fixes #584 Avoid JS error in role details view

### LINKED ISSUE #584: JS error in role details view
I registered an new privilege category provider for my entities. After that it is not possible to view or edit a role entity any more. The browser will raise a JavaScript error: **TypeError: categories.unshift is not a function** in **entity-category-tabs-component.js**.

There is an array expected but an object given. The PHP array used as input in tabsOptions.data looks like this:
```php
Array
(
    [0] => Array
        (
            [id] => account_management
            [label] => Account Management
        )

    [1] => Array
        (
            [id] => marketing
            [label] => Marketing
        )

    [2] => Array
        (
            [id] => sales_data
            [label] => Verkaufsdaten
        )

    [5] => Array
        (
            [id] => my_data
            [label] => My Bundle Entities
        )
)
```
The index value skips from 2 to 5, so the array is associative and json_encode will build an JS-object instead of JS-array.

digedag:
Since array_filter preserves keys it is necessary to call array_values after that. This will fix the issue.

### DIFF
--- a/src/Oro/Bundle/UserBundle/Provider/RolePrivilegeCategoryProvider.php
+++ b/src/Oro/Bundle/UserBundle/Provider/RolePrivilegeCategoryProvider.php
@@ -28,7 +28,7 @@ class RolePrivilegeCategoryProvider
     {
         $this->translator = $translator;
     }
-    
+
     /**
      * Add provider to registry
      *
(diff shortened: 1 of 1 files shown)


mkudelya:
Pull request has been merged.

Issue #492, opened 2016-07-06
Load datagrid.yml from multiple files


sbani:
Hi,

as our `datagrid.yml` file is growing and growing, we want to suggest a datagrid folder in `Rosources/config`. 

I thought about a combination of `FolderContentCumulativeLoader` and the `YamlCumulativeFileLoader`. One possibility is to add callback to the `FolderContentCumulativeLoader` which iterates over the loaded file names.

Example:

``` php
$folderLoader = new CumulativeConfigLoader(
    'oro_datagrid',
    [
        new YamlCumulativeFileLoader('Resources/config/datagrid.yml'),
        new FolderContentCumulativeLoader('Resources/config/datagrid/', -1, false, ['yml'], function($file) {
            $yamlParser = new YamlParser();
            try {
                return $yamlParser->parse(file_get_contents($file));
            } catch (ParseException $e) {
                $e->setParsedFile($file);
                throw new \InvalidArgumentException(sprintf('Unable to parse file "%s".', $file), $e->getCode(), $e);
            }
        }),
    ]
);
```

What do you think? Or do you know a better approach?

sergeyz:
I suggest it to store under `Resources/config/oro` folder

sadortun:
@sergeyz  `Ressources/config/oro/datagrid` ?

sergeyz:
Yep, to split from symfony configs

adamlundrigan:
The workflow bundle uses this notation to allow splitting up YAML files:

```yaml
imports:
    - { resource: 'b2b_flow_lead/steps.yml' }
    - { resource: 'b2b_flow_lead/attributes.yml' }
    - { resource: 'b2b_flow_lead/transitions.yml' }
    - { resource: 'b2b_flow_lead/transition_definitions.yml' }
```
([source](https://github.com/orocrm/platform/blob/2ac18cce58767af45462b78ddb569019782a4e0b/src/Oro/Bundle/WorkflowBundle/Resources/doc/reference/workflow/configuration-reference.md))

I tried the same approach with one of my `datagrid.yml` files but it does not work there ("A configuration for "xxxxxx" datagrid was not found.").  Perhaps that `imports` functionality can be generalized so it works in all the oro YAML files?

sbani:
In my opinion, @adamlundrigan's solution with `imports` is better than my first idea

StephenOTT:
Any movement on this??

mkudelya:
[...]

PR #589, opened 2016-12-17
Added  ability to import  yaml configuration  files recursively

dewos:
Hello there,
Related to #492.

This merge request allows to import recursively 'yml' config files using the standard Symfony notation for imports in yaml:

```php
# Resources/config/datagrid.yml

imports:
    - { resource: 'datagrid/user-grid.yml' }
    - { resource: 'datagrid/manager-grid.yml' }
    - { resource: 'datagrid/administrator-grid.yml' }
```

Note:

1. It works with any file loaded by a `YamlCumulativeFileLoader`
   - datagrid.yml
   - app.yml
   - navigation.yml
   - and so on....
2. An imported file can in turn import too, recursively.
3. Autocomplete for resources works too (with the PHPStorm plugin)
4. The change is "configuration agnostic" and should be compatible with any ORO version (1.10, 2.0). 
5. It is tested and I also added a Unit( @mkudelya  but I I do not have the time to update the docs right now :) )

### DIFF
--- a/src/Oro/Component/Config/Loader/YamlCumulativeFileLoader.php
+++ b/src/Oro/Component/Config/Loader/YamlCumulativeFileLoader.php
@@ -31,20 +31,20 @@ class YamlCumulativeFileLoader extends CumulativeFileLoader
 
             if (array_key_exists('imports', $configData) && is_array($configData['imports'])) {
                 if (empty($importedPaths)) {
-                    $importedPaths[] = $file; // for checking circular own-import
+                    $importedPaths[] = $file; // for checking circular self-import
                 }
 
+                $parent = new \SplFileInfo($file);
                 $imports = $configData['imports'];
                 unset($configData['imports']);
 
                 foreach ($imports as $importData) {
                     if (array_key_exists('resource', $importData)) {
-                        $parent = new \SplFileInfo($file);
                         $import = new \SplFileInfo($parent->getPath() . DIRECTORY_SEPARATOR . $importData['resource']);
                         $importPath = $import->getRealPath();
 
                         if (in_array($importPath, $importedPaths, true)) {
-                            $importedPaths[] = $importPath; // for a complete tree in the message
+                            $importedPaths[] = $importPath; // for a complete tree in the exception message
 
                             throw new \InvalidArgumentException(
                                 sprintf('Circular import detected in for "%s".', implode(' >> ', $importedPaths))
--- a/src/Oro/Component/Config/Loader/YamlCumulativeFileLoader.php
+++ b/src/Oro/Component/Config/Loader/YamlCumulativeFileLoader.php
@@ -19,14 +19,14 @@ class YamlCumulativeFileLoader extends CumulativeFileLoader
     /**
      * Parses a YAML file.
      *
-     * @param string $filePath Path to a file
+     * @param string $file Path to a file
      * @return array|null
      * @throws \InvalidArgumentException When loading of YAML file returns error
      */
-    protected function parseFile($filePath)
+    protected function parseFile($file)
     {
         try {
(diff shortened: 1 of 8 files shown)


src/Oro/Component/Config/Loader/YamlCumulativeFileLoader.php
```diff
@@ -18,18 +19,32 @@ protected function loadFile($file)
...
+
+            if (array_key_exists('imports', $configData) && is_array($configData['imports'])) {
+                $imports = $configData['imports'];
+                unset($configData['imports']);
+
+                foreach ($imports as $importData) {
+                    if (array_key_exists('resource', $importData)) {
+                        $file = new \SplFileInfo($filePath);
+                        $import = new \SplFileInfo($file->getPath() . DIRECTORY_SEPARATOR . $importData['resource']);
+                        $configData = array_merge_recursive($configData, $this->parseFile($import));
```
sadortun: Recursion is great, but you should add one of the following checks : 
- what is the maximum nesting level ? ( Ex throw after 255 includes )
- check if the file was already parsed `in_array ( $filePath, $this->alreadyParsedFiles)` ... throw inclusion loop detected

sadortun: Recursion is great, but you should add the following checks : 
- what is the maximum nesting level ? ( Ex throw after 255 includes )
- check if the file was already parsed `in_array ( $filePath, $this->alreadyParsedFiles)` ... throw inclusion loop detected

src/Oro/Component/Config/Tests/Unit/Loader/CumulativeConfigLoaderTest.php
```diff
@@ -277,4 +277,18 @@ public function testLoadEmptyFileWithoutWarnings()
 
         $this->assertEquals([$resource1], $result);
     }
+
```
sadortun: You should add the following test case:
```yaml
# myFile.yml
imports:
 - { resource: 'myFile.yml' }
```

And 
```yaml
# a.yml
imports:
 - { resource: 'b.yml' }

# b.yml
imports:
 - { resource: 'a.yml' }
```

src/Oro/Component/Config/Loader/YamlCumulativeFileLoader.php
dewos: @sadortun  Why? Maximum nesting level can lead to incongruent configuration, and infinite loops throws anyway during develping.

sadortun: It's better to throw early when a unrecoverable error is detected instead of running out of stack and memory !

src/Oro/Component/Config/Tests/Unit/Loader/CumulativeConfigLoaderTest.php
dewos: Done, thanks

src/Oro/Component/Config/Loader/YamlCumulativeFileLoader.php
dewos: Usually but not in this case. The maximum nesting level cannot be hardcoded, it is subjective (I MIGHT need 256 includes, or MORE....).

Adding a new argument in the class constructor will do, but it will be invasive, so for now I keep this as it is.

alsma: @dewos it's not required to be hardcoded, as @sadortun mentioned you can store a stack of currently processing and check if new import met is already in that stack then throw "circular import error"

mkudelya:
Hi @dewos 

I have merged PR to master and 1.10 branch.
Thanks!

PR #599, opened 2016-12-28
Fixes checking when using 0 as timestamp

roldan:


mkudelya:
Hi @roldan 

Thank you for your contribution. Your changes will be available in the few days.

PR #657, opened 2017-03-08
[Action Bundle]: Warm up actions.yml cache work very slowly

vtsykun:
| Q                | A
| ---------------- | -----
| Branch?| master|
|Bug fix?  | yes |
| New feature?| yes |
| BC Breaks? | no/yes |
| Deprecations? | no |
| Tests pass? | yes |
|License| MIT|
| Doc | yes |

### STR
As administrator i want to after update `actions.yml` cache will be rebuild automatically (on the fly) in debug mode. It's working now, but very slowly. 

1) Install application in dev mode with debug
2) Create `actions.yml` to display operation on any page
3) Go to this page to see this button
4) Update this `actions.yml`
5) Reload page

*Actual:* I have to wait for ~ 1 min when cache will be updated, because the all application cache, including `appDebugProjectContainer.php` is being updated, although the other resource files (such as  `services.yml` ) have not been changed.
*Expect:*  Only action cache should be updated

### Usage & Example 

The purpose of this PR is to update only cache that will be needed by the application without updating the cache of those resources, that have not been changed. This gives a better performance vs the old approach. Also it can be used to refactor [ContainerListener](https://github.com/orocrm/platform/blob/master/src/Oro/Bundle/DataGridBundle/EventListener/ContainerListener.php) in DatagridBundle

The following example shows how it can be used:

```yaml
# To register your config dumper:
oro.config.dumper:
    class: 'Oro\Component\Config\Dumper\CumulativeConfigMetadataDumper'
    public: false

# To register your config warmer with oro.config_cache_warmer.provider tag:
oro.configuration.provider.test:
    class: 'Acme\Provider\ConfigurationProvider'
    tags:
        - { name: oro.config_cache_warmer.provider, dumper: 'oro.config.dumper' }

```
```php
<?php

namespace Oro\Example\Dumper;

use Symfony\Component\DependencyInjection\ContainerBuilder;

use Oro\Component\Config\Dumper\ConfigMetadataDumperInterface;
use Oro\Bundle\CacheBundle\Provider\ConfigCacheWarmerInterface;
use Oro\Component\Config\CumulativeResource;

class ConfigurationProvider implements ConfigCacheWarmerInterface
{
    /**
    * @param ContainerBuilder $containerBuilder
    */
    public function warmUpResourceCache(ContainerBuilder $containerBuilder)
    {
        // some logic
        $resource = new CumulativeResource();
        $containerBuilder->addResource($resource);
    }
}

```


mkudelya:
Hi @vtsykun 

Would you please sign our Contributor License Agreement?
Thank you!

vtsykun:
Ok, I just now to accept the Contributor License Agreement https://www.orocrm.com/contributor-license-agreement

src/Oro/Bundle/CacheBundle/DependencyInjection/Compiler/CacheWarmerPass.php
```diff
@@ -0,0 +1,106 @@
...
+        $this->assertInstanceOf($definition, $container, ConfigMetadataDumperInterface::class);
+        $this->prepareDefinition($definition);
+    }
+
+    /**
+     * @param Definition $definition
+     */
+    protected function prepareDefinition(Definition $definition)
+    {
+        $definition->setShared(false)->setPublic(false);
```
ishirko: Looks like this will brake command 'oro:debug:operation'

ishirko:
If install OroCommerce Application (https://github.com/orocommerce/orocommerce) and apply changes from this pull request - some functional tests will fail with the error:

> Error message: Property Proxies\__CG__\Extend\Entity\EV_Prod_Inventory_Status::$id does not exist

src/Oro/Bundle/CacheBundle/DependencyInjection/Compiler/CacheWarmerPass.php
vtsykun: Fixed

vtsykun:
Hi, @ishirko  I was able to reproduce this bug, it occurs during loading the demo data loading Oro\Bundle\RFPBundle\Migrations\Data\Demo\ORM\LoadRequestWorkflowDemoData. Its cause in the`@duplicate` action what uses DeepCopy, if the original entity is instanseof doctrine proxy then the error occurs #https://github.com/myclabs/DeepCopy/issues/62 
Could you provide more information about failed tests (the classname of test what was failed). I'll think about how to fix it.
Thanks

ishirko:
Hi @vtsykun ,
Yes, you are right - mainly issues with proxies as I see. There is some failed functional tests:
1) ```
Oro\Bundle\OrderBundle\Tests\Functional\Controller\OrderPaymentCaptureTest::testOrderViewPageForUserWithPermission
Failed asserting that 500 matches expected 200.

2) Oro\Bundle\OrderBundle\Tests\Functional\Controller\OrderPaymentCaptureTest::testOrderViewPagePaymentTransactionSectionNotVisible
Failed asserting that 500 matches expected 200.

3) Oro\Bundle\SaleBundle\Tests\Functional\Workflow\QuoteBackofficeDefaultWorkflowTest::testCloneFromDraftStep
Failed asserting that 500 matches expected 200. Error message: Property Proxies\__CG__\Extend\Entity\EV_Prod_Inventory_Status::$id does not exist

4) Oro\Bundle\RFPBundle\Tests\Functional\Workflow\RfqFrontofficeDefaultWorkflowTestCase::testResubmitTransition
Failed asserting that 500 matches expected 200. Error message: Property Proxies\__CG__\Extend\Entity\EV_Prod_Inventory_Status::$id does not exist
```

Thank you

**Comment by vtsykun** [comment]
Hi, @ishirko I found the source of the problem and will fix it in short time.
Thanks!

**Comment by vtsykun** [comment]
Hi, @ishirko 
I fixed problem with failed functional tests. In the action config is allowed to use the parameters from ParameterBag. Like: 
```yml
- '@duplicate':
    settings:
        - [[keep], [propertyType, ['%oro_rfp.entity.request_product_item.class%']]]
```
This possibility is not documented and has not been used before, so I did not know it.

Thanks!
Vladimir

vtsykun:
I think that we can close the PR because there is no activity for it. Improving performance for updating cache only affects in *dev* mode and useful only to developers. For example according to blackfire profiles update cache for actions.yml more quickly by 90% from 1 min 2 s → 6.76 s. So I've moved all modification from this PR in a composer installable package https://github.com/vtsykun/oro-warmup-cache, so that other developers can use it.

src/Oro/Bundle/ActionBundle/Resources/config/cache.yml
```diff
@@ -32,3 +32,11 @@ services:
             - [addProvider, ['@oro_action.configuration.provider.action_groups']]
         tags:
             - { name: kernel.cache_warmer }
+
+    oro_action.config.dumper:
```
encoder32: These dumpers are used only in debug mode.
I think it will be better to register them in DIC only if debug mode is enabled.
Here is an example: https://github.com/laboro/dev/blob/master/package/platform/src/Oro/Bundle/ApiBundle/DependencyInjection/OroApiExtension.php#L74-L75

src/Oro/Bundle/CacheBundle/DependencyInjection/Compiler/CacheWarmerPass.php
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * @param Definition $definition
+     * @param ContainerBuilder $container
+     * @param string $class
+     */
+    protected function assertInstanceOf(Definition $definition, ContainerBuilder $container, $class)
+    {
+        $className = $definition->getClass();
+
+        if (!class_exists($className)) {
```
encoder32: I guess it can be safely replaced with
```
        if (0 === strpos($className, '%')) {
            $className = $container->getParameter(substr($className, 1, -1));
        }
```
what do you think?

src/Oro/Bundle/CacheBundle/DependencyInjection/Compiler/CacheWarmerPass.php
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * @param ContainerBuilder $container
+     * @param string $dumper
+     */
+    protected function assertDumper(ContainerBuilder $container, $dumper)
+    {
+        if (!$container->hasDefinition($dumper)) {
+            throw new \InvalidArgumentException('The declared cache metadata dumper "%s" not exist', $dumper);
```
encoder32: `sprintf` is missed

src/Oro/Bundle/CacheBundle/DependencyInjection/Compiler/CacheWarmerPass.php
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        $definition = $container->getDefinition($dumper);
+        $this->assertInstanceOf($definition, $container, ConfigMetadataDumperInterface::class);
+        $definition->setShared(false)->setPublic(false);
```
encoder32: do we really need this magic?

src/Oro/Bundle/CacheBundle/EventListener/CacheWarmerListener.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
+namespace Oro\Bundle\CacheBundle\EventListener;
+
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+use Oro\Component\Config\Dumper\ConfigMetadataDumperInterface;
+use Oro\Bundle\CacheBundle\Provider\ConfigCacheWarmerInterface;
+
+class CacheWarmerListener
```
encoder32: add a comment to describe the purpose of this listener

src/Oro/Bundle/CacheBundle/EventListener/CacheWarmerListener.php
```diff
@@ -0,0 +1,69 @@
...
+    ) {
+        $this->cacheMap->attach($configProvider, $dumper);
+    }
+
+    /**
+     * Check cache metadata and warm up cache if needed
+     */
+    public function onKernelRequest()
+    {
+        if (false === $this->kernelDebug) {
```
encoder32: I think it will be better to remove this check and register this service only if debug mode is enabled.
Here is an example: https://github.com/laboro/dev/blob/master/package/platform/src/Oro/Bundle/ApiBundle/DependencyInjection/OroApiExtension.php#L74-L75

src/Oro/Bundle/CacheBundle/Provider/ConfigCacheWarmerInterface.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Oro\Bundle\CacheBundle\Provider;
+
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+interface ConfigCacheWarmerInterface
```
encoder32: add a comment to describe the purpose of this interface

src/Oro/Bundle/CacheBundle/Provider/ConfigCacheWarmerInterface.php
```diff
@@ -0,0 +1,13 @@
...
+{
+    /**
+     * @param ContainerBuilder $containerBuilder
```
encoder32: add a comment to describe the what this method should do
it I understand correct this method should add resources that are used to build the given container builder in this container builder

src/Oro/Bundle/CacheBundle/Resources/config/services.yml
```diff
@@ -5,6 +5,13 @@ services:
     oro_cache.oro_data_cache_manager:
         class: %oro_cache.oro_data_cache_manager.class%
 
+    oro_cache.config_cache_warmer.listener:
+        class: 'Oro\Bundle\CacheBundle\EventListener\CacheWarmerListener'
+        arguments:
+            - '%kernel.debug%'
+        tags:
+          - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest }
```
vtsykun: It should also work for console commands, for example action groups can be used in workflow transitions which can be used in a cli process

src/Oro/Bundle/CacheBundle/EventListener/CacheWarmerListener.php
```diff
@@ -0,0 +1,69 @@
...
+        ConfigCacheWarmerInterface $configProvider,
+        ConfigMetadataDumperInterface $dumper
+    ) {
+        $this->cacheMap->attach($configProvider, $dumper);
+    }
+
+    /**
+     * Check cache metadata and warm up cache if needed
+     */
+    public function onKernelRequest()
```
vtsykun: For performance the method should be called only one times for master request.

src/Oro/Bundle/ActionBundle/Configuration/ConfigurationProvider.php
```diff
@@ -112,4 +150,28 @@ protected function resolveConfiguration(Collection $errors = null)
 
         return $data;
     }
+
+    /**
+     * @param ContainerBuilder|null $containerBuilder
+     * @return array
+     */
+    protected function loadConfiguration(ContainerBuilder $containerBuilder = null)
```
vtsykun: This code should be placed a separate service to avoid a injection of parameter bar, make easy for reuse and add local caching

src/Oro/Bundle/ActionBundle/Resources/config/cache.yml
vtsykun: Fixed

src/Oro/Bundle/CacheBundle/DependencyInjection/Compiler/CacheWarmerPass.php
vtsykun: I agree with you

PR #666, opened 2017-03-22
Fix #665 fatal error "Call to a member function getResizedImageUrl() on null"

### LINKED ISSUE #664: ReplaceNamespaces Migration fail
Hi Oro Team,

During the ReplaceNamespaces migration on 2.x, if the target audit field version already exists, the migration will stop and fail with an exception. There is proabaly a `try{} catch{}` missing in platform/src/Oro/Bundle/InstallerBundle/Migration/UpdateTableFieldQuery.php:98

Looking at the migration, the replacement is done via PHP code, instead of a single SQL query which would be more efficient.

If someone else have a similar issue, you can run the following queries on the oro_audit table to avoid duplicates:
```sql
update ignore `oro_audit` SET object_class=REPLACE(object_class,'OroCRM','Oro');
delete from oro_audit WHERE `object_class` LIKE 'OroCRM%';
```

Thanks
Samuel

### LINKED ISSUE #665: Call to a member function getResizedImageUrl() on null
Hi Oro Team

/cc @soloma88 

When viewing a Lead Actrivity Comment, 
```
Notice</b>:  Undefined property: Oro\Bundle\CommentBundle\Entity\Manager\CommentApiManager::$attachmentManager in /var/www/orocrm/vendor/oro/platform/src/Oro/Bundle/CommentBundle/Entity/Manager/CommentApiManager.php</b> on line <b>238</b><br />
Fatal error</b>:  Uncaught Error: Call to a member function getResizedImageUrl() on null in /var/www/orocrm/vendor/oro/platform/src/Oro/Bundle/CommentBundle/Entity/Manager/CommentApiManager.php:239
```

In  https://github.com/orocrm/platform/commit/96d100f4663c1e9a1e5848f0d1a05144f27fa793#diff-59732a79388412be1cf9270b3ccde35dL46

`$attachmentManager` was renamed to  `$attachmentProvider` but the reference https://github.com/orocrm/platform/blob/4c2271ce5c7bb37745bcc5dfd6083731a9741b67/src/Oro/Bundle/CommentBundle/Entity/Manager/CommentApiManager.php#L238
was left behind

sadortun:
Fix #665

Bug introduced in: 96d100f4663c1e9a1e5848f0d1a05144f27fa793

### DIFF
--- a/src/Oro/Bundle/CommentBundle/Entity/Manager/CommentApiManager.php
+++ b/src/Oro/Bundle/CommentBundle/Entity/Manager/CommentApiManager.php
@@ -13,6 +13,7 @@ use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 
 use Oro\Bundle\AttachmentBundle\Provider\AttachmentProvider;
+use Oro\Bundle\AttachmentBundle\Manager\AttachmentManager;
 use Oro\Bundle\CommentBundle\Entity\Comment;
 use Oro\Bundle\CommentBundle\Entity\Repository\CommentRepository;
 use Oro\Bundle\EntityBundle\Exception\InvalidEntityException;
@@ -44,6 +45,9 @@ class CommentApiManager extends ApiEntityManager
     /** @var AttachmentProvider */
     protected $attachmentProvider;
 
+    /** @var AttachmentManager */
+    protected $attachmentManager;
+
     /** @var aclHelper */
     protected $aclHelper;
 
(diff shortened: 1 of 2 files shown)


mkudelya:
Hi @sadortun 

The Oro team has already fixed this problem, but anyway thank you!

sadortun:
Ahhahahhaah,  please don't add `Open for Contribution` label for issues you already fixed !!

mkudelya:
I am awfully sorry for the mixup! At the time I did not know that the bug fix was in progress. Really sorry once again!

sadortun:
@mkudelya  No problem ! 👍

PR #667, opened 2017-03-24
Fix Composer constraints for the MessageQueueBundle bundle

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

As discussed on Slack with @sergeyz, I updated the Composer constraint for the Symfony package making it equal to the other bundle released standalone. Before merging, would it be possible to also allow Symfony `3.x`? I know the whole platform is not ready yet for it, but as this bundle is going to be used standalone and I don't think you're using anything that won't work on a newer Symfony version then it would be cool to allow people to install it on Symfony 3. In the worst case, a contribution can always be made to fix it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,9 @@ sudo: false
 language: php
 
 branches:
-  only:
-    - master
-    - /^\d+\.\d+$/
+    only:
+        - master
+        - /^\d+\.\d+$/
 
 php:
     - 5.6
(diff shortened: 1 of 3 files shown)


mbessolov:
@ste93cry can you actually test if it works with Symfony 3? We would rather not make assumptions and mislead the developers unless we know for sure that it works.

sadortun:
@ste93cry Travis testa should also probably run on both symfony 2.8 and 3.x even I the 3.x tests are allowed to fail to ensure the bundle remains compatible to both versions

ste93cry:
Should I then change the `.travis.yml` file to allow Symfony 3 tests for the whole platform? Otherwise I do not know how we could test only this bundle is compatible with that version of the framework

sbani:
I think it makes sense to test the whole platform because "why not". And makes sense to do so because running the newest symfony version isn't a big deal in *most* cases.

In context of the smaller/other vendors: Most of them implement features with less accuracy than symfony does. That leads to the fact, that these vendors put their effort on being compatible with the newest version of symfony only. In the end, if you want to use these vendor's newest features, you have to use the newest symfony.

ste93cry:
After a few try and retry I was finally able to get a Travis config that tests the whole platform with Symfony `2.8.13` and `3.x`. I have a few questions:

- As Symfony `3.0` is not supported anymore, should I remove it and only test `3.1` and `3.2`?
- Should I test the `dev-master` version of Symfony?
- Should I test any version of Symfony `2.8` even though `composer.json` specifies an exact version? Also, why the hell is Symfony locked up to `2.8.13` anyway?

About why the builds fails for Symfony 3.x versions, after investigating a bit there seems to be a few dependencies that in the version you require are not compatible. I don't know what do to, I would simply update them and check if everything still works to fix the problems

vbyndych:
Thank you for contribution. We have plans upgrade to Symfony 3.3, so it's probably an answer to your first and third questions. As for this PR it should be merged together with Symfony 3.3 upgrade, sorry for delay and thank you for your patience.
Internal ID: BAP-15236

anyt:
Upgrade to Symfony 3.4 already on the way. New version of MessageQueueBundle will be compatible only with Symfony `3.4.*` LTS.
ORO packages depends only to LTS versions of Symfony.
For previous LTS version of Symfony - `2.8.*` older versions of the bundle should be used.
Dependencies to Symfony will be upgraded for all the ORO packages at the same time to be compatible.
Thank you for initiative.

anyt:
Closed as not relevant anymore.

`oro/platform` 3.0.0 already compatible with Symfony 3.4.*.
MessageQueueBundle bundle as a separate package was abandoned, now it can be installed only as a part of the `oro/platform` package.

PR #672, opened 2017-04-06
Fix test that fail when there is a websocket listening

sadortun:
When a websocket server is listening on localhost:8080 theses tests will fail since a connection failure is expected


manowark:
@sadortun, thank you for the contribution. We will merge it in short time.

Issue #702, opened 2017-05-25
Datagrids: Support money values represented by integer


digedag:
It is a common pattern to [represent money values as integers](https://stackoverflow.com/a/3730040) to avoid precision issues on calculations. Symfony supports this pattern in money form type with a [divisor option](http://symfony.com/doc/current/reference/forms/types/money.html#divisor).

Oro datagrids should also support this kind of money values. I would like to add this two options:

```
datagrid:
    my-grid:
        columns:
            priceTotal:
                label:         mybundle.entity.price_total.label
                divisor:       100
                data_name:     priceTotal
                frontend_type: currency

        totals:
            grand_total:
                columns:
                    priceTotal:
                        formatter: currency
                        divisor:   100
                        expr:      'SUM(f.priceTotal)'
```

As you can see I would like to introduce the `divisor` option as Symfony does.

Right now I have some workarounds to display this kind of money values in datagrid. But I would like to see this feature in Oro-Core.

Would Oro-Team welcome this contribution?

PR #704, opened 2017-05-28
Datagrid divisor option

digedag:
This PR implements #702 . It makes it possible to handle money values as integers. With a new `divisor` option datagrid will always show the correct value.
This implementation supports formatter and totals extensions as well as filtering and inline editing.

If you need any changes please let me know!

### DIFF
--- a/src/Oro/Bundle/DataGridBundle/Resources/doc/backend/extensions/formatter.md
+++ b/src/Oro/Bundle/DataGridBundle/Resources/doc/backend/extensions/formatter.md
@@ -12,6 +12,7 @@ column_name:
     type: field # default value `field`, so this key could be skipped here
     frontend_type: date|datetime|decimal|integer|percent|currency|select|text|html|boolean # optional default string
     data_name: someAlias.someField # optional, key in result that should represent this field
+    divisor: some number # optional if you need to divide a numeric value by a number before rendering it
 ```
 Represents default data field.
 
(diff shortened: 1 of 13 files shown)


src/Oro/Bundle/DataGridBundle/Resources/doc/backend/extensions/formatter.md
```diff
@@ -12,6 +12,7 @@ column_name:
     type: field # default value `field`, so this key could be skipped here
     frontend_type: date|datetime|decimal|integer|percent|currency|select|text|html|boolean # optional default string
     data_name: someAlias.someField # optional, key in result that should represent this field
+    divisor: some number # optional if you need to divide a numeric value by a number before rendering it
```
sadortun: In https://github.com/orocrm/platform/pull/704/files#diff-ec92bcd45dd3ddbb7ca0081d1af0c23eR43 you use `100` as a default value, can you provide a quick example here as what kind of divisor value make sense ?

digedag: The value `100` will make sense for money if you want to store the amount as Cent (integer) into database. I think this is the most common use case. Then for instance there is `1042` persisted in database, and datagrid will show `10.42`, the value in Dollars.

manowark:
INTERNAL ID: BAP-15374

PR #733, opened 2017-08-01
BusinessUnitTreeTransformer. Fix error while save empty value 

vtsykun:
The next error occurred while save entity that have "owner_type"="BUSINESS_UNIT" without prefilled businiessUnit attribute
selection_074

### DIFF
--- a/src/Oro/Bundle/OrganizationBundle/Form/Transformer/BusinessUnitTreeTransformer.php
+++ b/src/Oro/Bundle/OrganizationBundle/Form/Transformer/BusinessUnitTreeTransformer.php
@@ -26,7 +26,7 @@ class BusinessUnitTreeTransformer implements DataTransformerInterface
     public function reverseTransform($value)
     {
         if (null === $value) {
-            return 0;
+            return null;
         } elseif (is_array($value)) {
             foreach ($value as &$val) {
                 if ($val === '') {


anyt:
Hi @vtsykun ,

Not sure this is a bug, as null to 0 was fix of another issue and I didn't found places in UI where it can be reproduced, seems you created some non standard form type for ownership configuration, in this case consider to use the standard one with allowClear=false, where user cannot create entities without owner

vtsykun:
The problem occurs in API. OwnerFormExtension add businessUnit fields for **all** forms for entity  "owner_type"=BUSINESS_UNIT. This broken POST and PATCH functionality of API. For example in UI datagrids with `inline_editing:true` uses PATCH method, but I did not find a place in the platform where you can see the error, because BUSINESS_UNIT not used often
I can provide next STR for with bug:

0) Create business unit BU1
1) Create entity with "owner_type"="BUSINESS_UNIT"
2) Create datagtid for this entity with `inline_editing:true`
*Note* For example you can use segment entity (Only needed update datagrids configuration)
```yml
datagrids:
    oro_segments-grid:
        inline_editing:
            enable: true
       .....
```
3) Create entity set owner BU1
4) Force remove business unit BU1 from database
5) Try to update value in datagrid
selection_081

src/Oro/Bundle/OrganizationBundle/Form/Transformer/BusinessUnitTreeTransformer.php
```diff
@@ -26,7 +26,7 @@ public function __construct($manager)
     public function reverseTransform($value)
     {
         if (null === $value) {
-            return 0;
```
vtsykun: I think this is typo error. Transformer should return value for model layer(BU entity or null), so integer type not allowed here.
Maybe here should be as 
```php
    public function reverseTransform($value)
    {
        if (null === $value) {
            $value = 0;
        } elseif (is_array($value)) {
```

manowark:
INTERNAL ID: BAP-15373

sergeyz:
I assume it was fixed in https://github.com/oroinc/platform/commit/0df2835a0e0f35eafbaa39fd2743402fc67cc81b

vtsykun:
Yes, Great thanks!

PR #775, opened 2017-12-10
Added the "dependencies" option in Resources/config/oro/bundles.yml

gplanchat:
In order to build bridge bundles, I added an option in the `Resources/config/oro/bundles.yml` files, therefore bridge bundles will only be activated when all (optional) dependencies are met.

## Use case

I have a `FooTimeTrackingBundle` working in OroPlatform. I want it to integrate with **OroCRM**, **OroCommerce** and **DiamanteDesk** when those bricks are available in my platform instance.

I don't want to make the dependencies hard and would like to make them soft, eg in the `suggest`section of the `composer.json` file:

```json
{
    "require": {
        "oro/platform": "2.*"
    },
    "suggest": {
        "oro/crm": "To activate customer information, OroCRM is required",
        "oro/commerce": "To activate invoicing of time spent, OroCommerce is required",
        "diamante/desk-application": "To activate customer care management, DiamanteDesk is required"
    }
}
```

## How it works

Let's suppose you have a `FooTimeTrackingBundle`, an autonomous bundle working in OroPlatform.

In order to make it work you will want to build bridge bundles: `FooTimeTrackingOroCRMBridge`, `FooTimeTrackingOroCommerceBridge`, `FooTimeTrackingDiamanteDeskBridge`.

Therefore, you will add this content to your bridge bundles in `Resources/config/oro/bundles.yml`:

```yml
bundles:
  - name: Foo\Bridge\TimeTrackingOroCRMBridge\FooTimeTrackingOroCRMBridge
    dependencies:
      - Oro\Bundle\AccountBundle\OroAccountBundle
      - Foo\Bundle\TimeTrackingBundle\FooTimeTrackingBundle
```

Soft dependency now loads your bridge bundle only when both `Oro\Bundle\AccountBundle\OroAccountBundle` and `Foo\Bundle\TimeTrackingBundle\FooTimeTrackingBundle` are activated

### DIFF
--- a/src/Oro/Bundle/DistributionBundle/OroKernel.php
+++ b/src/Oro/Bundle/DistributionBundle/OroKernel.php
@@ -140,7 +140,7 @@ abstract class OroKernel extends Kernel
             $import = Yaml::parse(file_get_contents($file));
             if (!empty($import)) {
                 if (!empty($import['bundles'])) {
-                    $bundles = array_merge($bundles, $this->getBundlesMapping($import['bundles']));
+                    $bundles = array_merge($bundles, $this->getBundlesMapping($import['bundles'], true));
                 }
                 if (!empty($import['exclusions'])) {
                     $exclusions = array_merge($exclusions, $this->getBundlesMapping($import['exclusions']));
@@ -152,15 +152,35 @@ abstract class OroKernel extends Kernel
 
         uasort($bundles, array($this, 'compareBundles'));
 
+        do {
+            $bundleNames = array_keys($bundles);
+            $removedBundles = array_filter($bundles, function ($config) use ($bundleNames) {
+                if (!isset($config['dependencies'])) {
+                    return false;
+                }
+
+                foreach ($config['dependencies'] as $dependency) {
+                    if (!in_array($dependency, $bundleNames)) {
+                        return true;
+                    }
+                }
+
+                return false;
+            }, ARRAY_FILTER_USE_BOTH);
+
+            $bundles = array_diff_key($bundles, $removedBundles);
+        } while (count($removedBundles) > 0);
+
         return $bundles;
     }
 
     /**
-     * @param $bundles
+     * @param array $bundles
+     * @param bool $withDependencies
      *
      * @return array
      */
-    protected function getBundlesMapping(array $bundles)
+    protected function getBundlesMapping(array $bundles, bool $withDependencies = false)
     {
         $result = array();
         foreach ($bundles as $bundle) {
@@ -180,6 +200,13 @@ abstract class OroKernel extends Kernel
                 'kernel'   => $kernel,
                 'priority' => $priority,
             );
+
+            if ($withDependencies === true && isset($bundle['dependencies']) && is_array($bundle['dependencies'])) {
+                $result[$class]['dependencies'] = [];
+                foreach ($bundle['dependencies'] as $dependency) {
+                    $result[$class]['dependencies'][] = (string) $dependency;
+                }
+            }
         }
 
         return $result;


src/Oro/Bundle/DistributionBundle/OroKernel.php
```diff
@@ -152,15 +152,35 @@ protected function collectBundles()
 
         uasort($bundles, array($this, 'compareBundles'));
 
+        do {
+            $bundleNames = array_keys($bundles);
+            $removedBundles = array_filter($bundles, function ($config) use ($bundleNames) {
+                if (!isset($config['dependencies'])) {
+                    return false;
+                }
+
```
mbessolov: @gplanchat  This doesn't look efficient as we are filtering out the same bundles without dependencies again and again in each loop.

gplanchat: Yes, thus there is a good reason for this : disabling a bundle can remove a dependency of another bundle and require it to be disabled on 2nd pass, etc.

This processing is made once, at cache warmup, it don't affect runtime

alexandr-parkhomenko: @gplanchat This solution has some big drawback, we can't lock required version of the package with the functionality `suggestion`. In case if suggested bundle changes own API we will receive broken functionality, I think you agree with me that this inappropriate for the system like OroPlatform.

I think better solution when we host bridge bundle separately from the monolithic repository and include it only when we need it. This bundle will have strong dependencies, so we know that if someone wants to use this functionality he receive worked solution.

mbessolov: @gplanchat  I understand why it is done, I'm just saying that it is not an optimal (algorithmically) way to do it.

gplanchat: @alexandr-parkhomenko I don't think there is a reasonable way to integrate composer in this case. The goal is to keep things simple.

3 things can help to maintain compatibility :
* Oro packages require specific OroPlatform version, which is a common dependency managed by Composer, your solution about declared dependencies in the `composer.json`of the bridge bundle actually does the version compatibility check and is not exclusive with this option
* SemVer guarantees to not break things at any time, if it is properly followed
* a `README.md` file with a version compatibility list could solve most cases

There are not thousands of Oro packages for now, most of the needs you are mentioning can be handled by composer (which is the best option). This change provide a way to tell Oro to not load the bundle in the kernel when all dependencies are not met.

@mbessolov maybe we could Blackfire it to find some better algorithm

gplanchat: @alexandr-parkhomenko let me describe you a broken use case happening on current versions:

In `FooBundle`
```yaml
bundles:
  - name: Foo\FooBundle
``` 

In `BarBundle`
```yaml
bundles:
  - name: Bar\BarBundle
``` 

In `FooBarBridgeBundle`
```yaml
bundles:
  - name: Foo\FooBarBridgeBundle
``` 

In `BazBundle`
```yaml
bundles:
  - name: Baz\BazBundle
exclusions:
  - name: Bar\BarBundle
``` 

The `BarBundle` will be excluded by the `BazBundle`, but not the bridge which will be loaded either by composer and the OroKernel.

gplanchat:
@mbessolov 
Initial version:
image

Modified version:
image

I ran several times the tests, no difference can be clearly measured, either in performance or in memory consumption.

gplanchat:
@mbessolov do you already have some updates on the changes we will see in Oro 3.0 concerning this part ?

anyt:
Hi @gplanchat,

3.0 Release Candidate already released. 
Now we are working on testing and stabilization of 3.0 LTS. 
New features should not be added to this release as we'll not have time to test them deeply, and we can't release untested functionality.

mbessolov:
@gplanchat  We'll work on enabling Symfony Flex for 3.1. Considering that Flex automatically enables the installed bundles, this PR should be reworked.

PR #800, opened 2018-01-30
Fix bad reference to $currentUser parameter

curiosity26:
$currentUser was being used as a dynamic variable name on $this. Needs to be referenced directly as a User object.


aalgogiver:
Thank you. Internal task id is BAP-16481.

manowark:
Your PR has been merged. Changes from PR will be synced to the public repository in a few days. Thank you.

PR #844, opened 2018-05-29
EntityExtendBundle: add ORM reflection method

gplanchat:
Added a way to get the column name from the class name and field name. When using enums and loads of EntityExtend migrations, it is not easy to find out the column name

### DIFF
--- a/src/Oro/Bundle/EntityExtendBundle/Migration/Extension/ExtendExtension.php
+++ b/src/Oro/Bundle/EntityExtendBundle/Migration/Extension/ExtendExtension.php
@@ -905,6 +905,19 @@ class ExtendExtension implements NameGeneratorAwareInterface
     }
 
     /**
+     * Gets a field name by a table name and a column name
+     *
+     * @param string $className
+     * @param string $fieldName
+     *
+     * @return string|null
+     */
+    public function getColumnNameByFieldName($className, $fieldName)
+    {
+        return $this->entityMetadataHelper->getColumnNameByFieldName($className, $fieldName);
+    }
+
+    /**
      * @param Table|string $table A Table object or table name
      *
      * @return string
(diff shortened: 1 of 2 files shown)


yaroslavbr:
Hi @gplanchat ,

Thank you for your PR! We greatly appreciate all contributions to Oro projects.

Internal task id #BAP-17148

src/Oro/Bundle/EntityExtendBundle/Migration/Extension/ExtendExtension.php
```diff
@@ -904,6 +904,19 @@ public function getFieldNameByColumnName($tableName, $columnName)
         return $this->entityMetadataHelper->getFieldNameByColumnName($tableName, $columnName);
     }
 
+    /**
+     * Gets a field name by a table name and a column name
```
anyt: You have `table name` in the comment, but `class name` in an argument name.

src/Oro/Bundle/EntityExtendBundle/Migration/Extension/ExtendExtension.php
```diff
@@ -904,6 +904,19 @@ public function getFieldNameByColumnName($tableName, $columnName)
...
+     *
+     * @param string $className
+     * @param string $fieldName
+     *
+     * @return string|null
+     */
+    public function getColumnNameByFieldName($className, $fieldName)
```
anyt: Can you show an example when this method is useful?
for me, it can make some mess in migrations as currently we always work there with table names and column names, but with this method, there will be mix with class names and field names. It looks less readable.

gplanchat: When adding enums, you can't get the field name :

```
$this->extendExtension->addEnumField(...)
```

In fact, it is using the 3rd argument + `_id`

There is no method for removing an enum nor any method to build this relation field name, only line in the code. Should we assume it is a convention ?

PR #854, opened 2018-07-30
Cache clear twice as fast

vtsykun:
### Benchmark test

| command                                               | Patch kernel (sec.) |    Old kernel (sec.) |
|-------------------------------------------------------|---------------------|----------------------|
| `rm -rf var/cache && php bin/console --env=dev`       | 9.25                |  16.11               |
| `rm -rf var/cache && php bin/console --env=prod`      | 7.64                |  14.46               |

Hi. During the booting bundle [OroEntityExtendBundle](https://github.com/oroinc/platform/blob/master/src/Oro/Bundle/EntityExtendBundle/OroEntityExtendBundle.php#L60) the separate process is created to check extend entity cache, so the container will be compiled and dumped twice: in the main process and the its child process. These changes allow you to clear cache 2x times faster.


anyt:
Wow! 
Thank you for contribution @vtsykun. 
Need to check it first, but looks promising.

dewos:
@anyt any news about this one? Thanks

anyt:
I've checked that PR, but not sure about the current solution, as it covers only one case when the application will try to build the container cache multiple times in parallel, I think at the core we should have the more universal solution. Maybe using some locking mechanism so other processes will wait till cache will be built or something like that.

Don't have time yet to investigate other approaches. I plan to allocate more time for that at the next month.

PR #752, opened 2017-09-18
Fix execution of js-scripts on inline embedded forms

scapital:
One of option of embed of [EmbeddedForm](https://github.com/oroinc/platform/tree/2.3/src/Oro/Bundle/EmbeddedFormBundle) is `inline`.

If a view of form type or a view of layout have some javascript, it isn't executing, because a browser don't know what code is javascript.

The PR fixes it.

Thanks!

PR #861, opened 2018-09-05
Fix execution of js-scripts on inline embedded forms 2

vtsykun:
Additional fixes for #752.
By default the browser for `document.createElement('script')` loads the JavaScript asynchronously, in this PR was fix inherit of `async`

### DIFF
--- a/src/Oro/Bundle/EmbeddedFormBundle/Resources/public/js/embed.form.js
+++ b/src/Oro/Bundle/EmbeddedFormBundle/Resources/public/js/embed.form.js
@@ -79,7 +79,8 @@ var ORO = (function(ORO) {
                     realTag = document.createElement('script');
 
                     if (scripts[k].hasAttribute('src')) {
-                        realTag.setAttribute('src', scripts[k].getAttribute('src'));
+                        realTag.async = scripts[k].async;
+                        realTag.src = scripts[k].src;
                     } else {
                         realTag.innerHTML = scripts[k].innerHTML;
                     }


src/Oro/Bundle/EmbeddedFormBundle/Resources/public/js/embed.form.js
```diff
@@ -79,7 +79,8 @@ var ORO = (function(ORO) {
                     realTag = document.createElement('script');
 
                     if (scripts[k].hasAttribute('src')) {
-                        realTag.setAttribute('src', scripts[k].getAttribute('src'));
+                        realTag.async = scripts[k].async;
+                        realTag.src = scripts[k].src;
                     } else {
                         realTag.innerHTML = scripts[k].innerHTML;
                     }
```
Grygir: Let's improve a bit the solution to cover the case if some other attribute is required in `realTag` (e.g. `type` or `id`):
```js
    for (var i = 0; i < scripts[k].attributes.length; i++) {
        realTag.setAttribute(scripts[k].attributes.name, scripts[k].attributes.value);
    }

    if (!scripts[k].hasAttribute('src')) {
        realTag.innerHTML = scripts[k].innerHTML;
    }
```
And let's copy attribute even if it's inline script (e.g. in case the inline script has `type="text/template"`)

Thanks.

anyt: Also attribute copying should be placed outside `if (scripts[k].hasAttribute('src')) {` 
statement as inline scripts also can have attributes

vtsykun: Good idea :+1: , it looks better, I fixed it in PR
Thanks

src/Oro/Bundle/EmbeddedFormBundle/Resources/public/js/embed.form.js
```diff
@@ -77,10 +77,11 @@ var ORO = (function(ORO) {
...
 
                     realTag = document.createElement('script');
+                    for (var i = 0; i < scripts[k].attributes.length; i++) {
```
vtsykun: Element `<script src="">` may not have `async` attribute, so `scripts[k].hasAttribute('async')` return false. Without  `async` attribute the script tag load synchronously, this ensures that scripts are executed in correct order, but `document.createElement('script')` loads the JS asynchronously by default. To it fix needs also add a new checks

```js
if (!scripts[k].hasAttribute('async')) {
    realTag.async = false;
}
```

Grygir: Looks like you are right. Have checked the issue here https://jsfiddle.net/hoybuLej/42/

Let's add code that you have proposed.

PR #864, opened 2018-09-21
Properties instead of properies

startupz:
Typographical error


anyt:
Merged,
Thank you for the contribution!

PR #875, opened 2018-12-04
Fix Parent Service Definition

RocKordier:


src/Oro/Bundle/DataGridBundle/Resources/config/mass_actions.yml
```diff
@@ -82,7 +82,7 @@ services:
 
     oro_datagrid.extension.mass_action.type.massedit:
         class: Oro\Bundle\DataGridBundle\Extension\MassAction\Actions\Widget\MassEdit\MassEditAction
-        parrent: oro_datagrid.extension.mass_action.type.window
+        parent: oro_datagrid.extension.mass_action.type.window
```
anyt: This definition is redundant as there is nothing to extend from `oro_datagrid.extension.mass_action.type.window` service it makes sense to delete parent definition at all.

RocKordier: Hey @anyt ,
okay, i doesn't looked deep into the code. 
I've removed the `parent` statement and force pushed the change.

Thx Eric

RocKordier: Nope, unfortunately not.
But I already discussed this kind of tool with @Baachi

I've found it in the profiler logs
screenshot 2018-12-11 at 17 18 59

anyt: Logs are enough to automate this check, thank you.

RocKordier:
Committed in 2a14792824264cd7e437349a9ba52f840a903d37 
thx.
closed

PR #879, opened 2018-12-07
Better handling for file downloads

RocKordier:
* make it possible to download files in Safari
* Guess mimetype via gaufrette adapter

With the actual implementation it is very annoying to download files with safari. The mime type is not handled correct and the download prompt doesn't pop up. That is fixed now.

Thx Eric


RocKordier:
Merged,
Merry Christmas 🎅
Eric

PR #881, opened 2018-12-18
Datagrid filters missing after deleting an entity

vtsykun:
When using any filter on datagrid and then going to detail view of any record and deleting it they are redirected back to datagrid view but filters are lost.


anyt:
*Internal  ticket id #BAP-18136*

PR #884, opened 2019-01-07
[MessageQueue] Allow all callables not only closures

Baachi:
Hey!

the job runner does actually only supports `Closures` instead of any `callable`. Closures are generally bad to (unit)test. So i try to avoid to use this pattern.

PHP does support the `callable` typehint which provides a lot more flexibility.

### DIFF
--- a/src/Oro/Component/MessageQueue/Job/JobRunner.php
+++ b/src/Oro/Component/MessageQueue/Job/JobRunner.php
@@ -40,7 +40,7 @@ class JobRunner
      *
      * @return mixed
      */
-    public function runUnique($ownerId, $name, \Closure $runCallback)
+    public function runUnique($ownerId, $name, callable $runCallback)
     {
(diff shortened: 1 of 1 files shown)


anyt:
Hi @Baachi ,

I've discussed your proposal with the message queue component lead. And for now, we don't want to allow using class methods here. 

Enabling `callable` will allow developers to move logic outside of the processor and could make the code less obvious and readable. So we decided to stay with the Closure typehint.

The goal is to make this Closure very simple and don't have any complex logic, so usually we don't cover it with unit tests and just mock it. 
Here is an example of a test that we usualy have on it
https://github.com/oroinc/platform/blob/master/src/Oro/Bundle/SearchBundle/Tests/Unit/Async/ReindexEntityMessageProcessorTest.php#L53-L57

If you have an example when unit testing of that method could be helpful, please provide it and we will duscuss.

Thank you for you contribution.
Best regards!

PR #888, opened 2019-01-16
Update composer.json

iklymenko:


anyt:
composer.json should be updated from the monolithic repository only.

PR #892, opened 2019-02-11
Fixed an exception while using Postgres on 3.1.2

gplanchat:
Got error due to Postgres requiring a call to CAST(...)
`SQLSTATE[XX000]: Internal error: 7 ERROR: failed to find conversion function from unknown to text").`


anyt:
Hi @gplanchat,
The circumstances when the fix is needed is quite hard to understand for me. 
 
Can you share the full query or the query builder, where the fix was required?

PR #902, opened 2019-03-22
Add Webpack Progress flag to assets:build command

chrisaligent:
When using `php bin/console oro:assets:build --watch`, when files are modified nothing is displayed in the terminal output (even though the builds are running successfully). This makes it difficult to tell if Webpack has detected your code changes, or if it's actually building the assets.

Webpack has a handy `--progress` CLI parameter to watch the progress of builds:
https://webpack.js.org/api/cli/#debug-options

Adding the `--progress` parameter allows you to see a progress percentage, along with the filename currently being built.
The output looks something like this:
```
[122] ./css/layout/basetheme/styles.js 28 bytes {11} [built]
+ 221 hidden modules
65% building 220/228 modules 8 active ...public/css/layout/blank/styles.css.scss 
```

This Pull Request adds `--progress` as an optional flag, including usage instructions.

Another possible future enhancement could be to allow passing additional flags to the Webpack CLI, something like:
```
php bin/console oro:assets:install --watch --webpack-opts='--debug --verbose --bail --cache=false'
```

This would allow devs fine-grained control over the way the build runs, which can be important in CI environments.

### DIFF
--- a/src/Oro/Bundle/AssetBundle/Command/OroAssetsBuildCommand.php
+++ b/src/Oro/Bundle/AssetBundle/Command/OroAssetsBuildCommand.php
@@ -100,6 +100,12 @@ DESCRIPTION
                 webpack continues to watch the changes in any of the resolved files.'
             )
             ->addOption(
+                'progress',
+                'p',
+                InputOption::VALUE_NONE,
+                'Show build progress as a percentage in output (verbose mode).'
+            )
+            ->addOption(
                 'npm-install',
                 'i',
                 InputOption::VALUE_NONE,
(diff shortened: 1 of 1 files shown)


chrisaligent:
(I just signed the CLA, rather than waiting for the `orocla` bot to post the link).

anyt:
What do you think about enabling progress by default? 
It's a common practice to show progress by default, as I see. And I can't find any scenarios when you need to disable progress.

chrisaligent:
I like that idea, however because of the way it writes the output to the
terminal I worry that it could interfere with CI/CD output if enabled by
default.

The standard workaround for tools like this (eg Yarn) is to detect the "CI"
environment variable and automatically disable this kind of output. Same
thing with the --color option too I think.

As far as I know the majority of CI platforms have implemented the "CI"
environment variable (Travis, Jenkins, Gitlab).

That keeps CI logs clean.

I haven't checked if webpack supports CI mode already.

PR #908, opened 2019-04-05
Fixed route name for delete operation

mchojrin:


src/Oro/Bundle/ActionBundle/Resources/doc/operations.md
```diff
@@ -149,7 +149,7 @@ php bin/console oro:action:configuration:validate
 **Oro Action Bundle** defines several system wide default operations for a common purpose. Those are basic CRUD-called operations for entities:
  
  - `UPDATE` - operation for an entity editing that uses a route from the `routeUpdate` option of the entity configuration.
- - `DELETE` - operation for an entity deletion that uses a route from the `routeName` option of the entity configuration.
```
aivus: Hi @mchojrin 

Why do you think that `routeDelete` should be here?

As I see it uses `routeName` config value for DELETE operation:
https://github.com/oroinc/platform/blob/3f34fe6586618288d9de9bb10ba5fff7668a75dd/src/Oro/Bundle/ActionBundle/Resources/config/oro/actions.yml#L75

mchojrin: @aivus ok... maybe I was too quick with the suggestion :p

Sorry.

anyt:
Closed, as nothing to do here.

PR #915, opened 2019-04-25
Fix the oro:migration:dump command for unprefixed bundles

gplanchat:
When using unprefixed bundles, eg. `AppBundle`, the `namespace` property gets filled with the value `Entity\Entity` instead of `AppBundle\Entity`


anyt:
Thank you for your contribution!
_Internal ticket id #BAP-18773_

PR #920, opened 2019-05-26
Fix issue #919 tag name are wrapped when contains space

### LINKED ISSUE #919: Tags with multiple words are broken down
Hi Oro Team,

Small UI bug that is quite anoying :( 

When tags contains multiple words, the words are broken down on multiple lines
For example : 

image

image

Seems like a simple addition to CSS can fix the issue:
```css
.tags-container__tag-entry {
    white-space: nowrap;
}
```

Have a good day,
Samuel

sadortun:
Fix #919


anyt:
Fixed in a different way. See https://github.com/oroinc/platform/issues/919#issuecomment-496883531.

PR #924, opened 2019-05-27
Fix issue #923 Segment Filter: Multi-Select 'not-any-of' returns results from 'any-of'

### LINKED ISSUE #923: Segment Filter: Multi-Select "Not any of" not working
Hi Oro Team

The multi-select "not any of" filter does not work anymore in 3.1.

How to reproduce

1) I've added an multi-enum field to the Contacts  `testenum`

image

2) I've created a test segment including the field and values to exclude: 
image

3) Results that are displayed are wrong.
```sql
SELECT o0_.id AS id_0, o1_.email AS email_1, o0_.testenum_ss AS testenum_ss_2, o0_.id AS id_3, o0_.organization_id AS sclr_4, o0_.user_owner_id AS sclr_5 
FROM orocrm_contact o0_ 
LEFT JOIN orocrm_contact_email o1_ ON o0_.id = o1_.owner_id AND (o1_.is_primary = true) WHERE o0_.id IN (SELECT o2_.id FROM orocrm_contact o2_ 
LEFT JOIN orocrm_contact_email o3_ ON o2_.id = o3_.owner_id AND (o3_.is_primary = true) WHERE o2_.id IN (SELECT o4_.id FROM orocrm_contact o4_ 
INNER JOIN oro_rel_83dfdfa4bc56bacb3aa86b o6_ ON o4_.id = o6_.contact_id 
INNER JOIN oro_enum_dee077b0_f846c8a o5_ ON o5_.id = o6_.ev_contact_testenum_f846c8a_id WHERE o5_.id IN ('a', 'b', 'c')) AND o2_.organization_id = 1) AND o0_.organization_id = 1 ORDER BY o0_.id ASC LIMIT 25 OFFSET 0
```

Confirmed by the SQL above `WHERE o5_.id IN ('a', 'b', 'c'))`
image

PS: Also tested with TAGS, and Tags `not any of` is _working as expected_

Have a good day,
Samuel

sadortun:
Fix #923 

This PR fix the issue, BUT you should investigate WHY the `$comparaisonType` is no longer a number. I did not cast it directly since i have no idea if it can be something else than a number.

Merge with caution :)

### DIFF
--- a/src/Oro/Bundle/FilterBundle/Filter/MultiEnumFilter.php
+++ b/src/Oro/Bundle/FilterBundle/Filter/MultiEnumFilter.php
@@ -79,7 +79,7 @@ class MultiEnumFilter extends BaseMultiChoiceFilter
             $fieldName,
             $parameterName,
             $this->getName(),
-            in_array($comparisonType, [DictionaryFilterType::NOT_EQUAL, DictionaryFilterType::TYPE_NOT_IN], true)
+            in_array($comparisonType, [DictionaryFilterType::NOT_EQUAL, DictionaryFilterType::TYPE_NOT_IN], false)
         );
     }
 


mbessolov:
_Internal ID: BAP-18936_

anyt:
Fixed in a different commit https://github.com/oroinc/platform/commit/32e028ec4468f074b16b77271da5c7212da1a852.
Thank you for the contribution, your solution was a good starting point for the team!

PR #952, opened 2019-09-23
WorflowEntityValidator allow object

vytsci:
If entity contains some object rather than entity and it goes to validator it throws not manageable error. We have lots of DataTypes in our code. This is just an unnecessary restriction.

### DIFF
--- a/src/Oro/Bundle/WorkflowBundle/Validator/Constraints/WorkflowEntityValidator.php
+++ b/src/Oro/Bundle/WorkflowBundle/Validator/Constraints/WorkflowEntityValidator.php
@@ -198,7 +198,9 @@ class WorkflowEntityValidator extends ConstraintValidator
     {
         $fieldValue = $this->propertyAccessor->getValue($object, $restriction['field']);
         if (is_object($fieldValue)) {
-            $fieldValue = $this->doctrineHelper->getSingleEntityIdentifier($fieldValue);
+            $fieldValue = $this->doctrineHelper->isManageableEntity($fieldValue) ?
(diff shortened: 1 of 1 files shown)


mbessolov:
@vytsci Workflow engine in OroPlatform is built for transitioning *entities* through different steps of a workflow, and the *entity* restrictions are also for *entities* - https://github.com/oroinc/platform/blob/master/src/Oro/Bundle/WorkflowBundle/Resources/doc/reference/workflow/configuration-reference.md#entity-restrictions-configuration

Why would there ever be a non-entity object?

vytsci:
Issue is that we have status as an object, but it can be managed as a string, and ORO thinks that if you have property and it is an object it must be an entity and it should not be like that. I understand that relations must be entities, but for example DateTime is also an object and if you add that to this equation it will also throw non configurable entity error.

src/Oro/Bundle/WorkflowBundle/Validator/Constraints/WorkflowEntityValidator.php
```diff
@@ -198,7 +198,9 @@ protected function validateAllowedValues($object, $message, $restriction)
     {
         $fieldValue = $this->propertyAccessor->getValue($object, $restriction['field']);
         if (is_object($fieldValue)) {
-            $fieldValue = $this->doctrineHelper->getSingleEntityIdentifier($fieldValue);
+            $fieldValue = $this->doctrineHelper->isManageableEntity($fieldValue) ?
+                $this->doctrineHelper->getSingleEntityIdentifier($fieldValue) :
+                (string)$fieldValue;
```
anyt: What if the related object doesn't implement the `__toString()` method?

PR #961, opened 2019-10-17
Fix incorect template path

sadortun:


anyt:
Hi @sadortun,

Please explain why the path is incorrect?
Without these changes, different templates for different widget types, like dialog, were selected automatically. But now it doesn't work.

sadortun:
Hi @anything

First it seems this controller is not used from any oro code. ( searching by route name) 

Also for the controller to find the path automatically it need to be under

Ressources/views/Grid/widget.html.twig

But the file is currently under

Ressources/views/Grid/widget/widget.html.twig

Can you please explain how the following works ? 
> Without these changes, different templates for different widget types, like dialog, were selected automatically

vtsykun:
Hi @sadortun 
> Also for the controller to find the path automatically it need to be under

There is exist listener in [UIBundle](https://github.com/oroinc/platform/blob/3.1/src/Oro/Bundle/UIBundle/EventListener/TemplateListener.php#L39) to select different templates for different widget types.
For example
1) https://demo.orocrm.com/index.php/datagrid/widget/users-grid?_widgetContainer=dialog - render grid for [dialog widget](https://github.com/oroinc/platform/blob/3.1/src/Oro/Bundle/DataGridBundle/Resources/views/Grid/dialog/widget.html.twig)
2) https://demo.orocrm.com/index.php/datagrid/widget/users-grid?_widgetContainer=block - render [default grid template](https://github.com/oroinc/platform/blob/3.1/src/Oro/Bundle/DataGridBundle/Resources/views/Grid/dialog/widget.html.twig)

If you need to load the grid with ajax, you can add get parameter `_widgetContainer=block` into request

anyt:
I'm closing the PR as there is nothing wrong with the template path.
Thank you!

sadortun:
@vtsykun Awesome, thanks for the details !

Issue #925, opened 2019-06-05
Querybuilder / Segment SQL Optimisation


sadortun:
Hi Oro Team,

We are investigating why on some query the Segment page become very slow ( 30-45 seconds to load with complex query),

# Topic 1 
We found that the filter query is executed twice and the only difference is the `LIMIT` clause. and there is query executed multiple times

 - If you look bellow at query # 57 and query # 59.  Is # 57 nessessary ? If not, this could improve loading time by 25% ! 
- If you look at query # 56 and # 58, they are **identical** .... That would be another 25% improvement in speed.

Optimizing theses two would give users a 50% improvement in all features related to segments ! 

(example with query   `email contains "a"`
![image](https://user-images.githubusercontent.com/794449/58928185-88c72800-871f-11e9-93de-d875fe0e39e7.png)

# Topic 2

In our use case, the COUNT() query takes almost **15 seconds**, compared to the SELECT which takes about **300ms**. 

- Do you have any idea why the COUNT() is taking such excruciating (**50 times**) time to run ?

![image](https://user-images.githubusercontent.com/794449/58929296-54a23600-8724-11e9-85b4-ae15cb030eb3.png)

Have a good day,
Samuel

sadortun:
The query # 57 seems to come from  https://github.com/oroinc/platform/blob/master/src/Oro/Bundle/QueryDesignerBundle/Validator/QueryValidator.php#L105

Is it normal the validation run every time the segment is displayed ?

sadortun:
After some troubleshooting, i've found out that the query are only executed by the validate() method

https://github.com/oroinc/platform/blob/master/src/Oro/Bundle/QueryDesignerBundle/Validator/QueryValidator.php#L100

And results from query # 56 and # 57 are discarded. Then re-executed by the appropriate getGridData()

Is it normal the validation run every time the segment is used ?

aivus:
Hi @sadortun 
Thank you for your investigation

We need some time to check this

sadortun:
Hi @aivus 

Do you have an update on this ? Can you provide an ETA ? It's a blocking issue for us. 

We are willing to pay to get support on this issue quickly if it can help to get a faster resolution.

Thanks,
Samuel

anyt:
Hi @sadortun,
[...]

PR #963, opened 2019-10-17
Load segment data asynchronously

sadortun:
Segment QueryFilter can be quite complex and long to execute, this change allow to load the datagrid async, therefore allowing the page to load quickly

Require fix by #961
Related to #925

I am mostly sure that you wont merge this PR as is, but it give you a starting point to fix this issue

### DIFF
--- a/src/Oro/Bundle/SegmentBundle/Resources/views/Segment/view.html.twig
+++ b/src/Oro/Bundle/SegmentBundle/Resources/views/Segment/view.html.twig
@@ -54,8 +54,27 @@
             {% set gridParams = gridParams|merge({'display_sql_query': true}) %}
         {% endif %}
         {% set params = params|default({})|merge(gridParams) %}
-        {{ dataGrid.renderGrid(gridName, params, renderParams) }}
-    {% else %}
+        <div id="async-datagrid">
+            {% include "OroUIBundle::view_loading.html.twig" %}
+        </div>
+
+        <script type="text/javascript">
+
+          require(['jquery', 'oroui/js/app/components/component-manager'],
+            function ($, ComponentManager) {
+              var $asyncDatagrid = $('#async-datagrid')
+
+              $.get('{{ path('oro_datagrid_widget', {gridName: gridName} ) }}', {
+                  params: {{ params | json_encode|raw }},
+                  renderParams: {{ renderParams |json_encode |raw }}
+                },
+                function (data) {
+                  $asyncDatagrid.html(data)
+                  var manager = new ComponentManager($asyncDatagrid)
+                  manager.init()
+                })
+            })
+        </script>    {% else %}
         <div class="container-fluid">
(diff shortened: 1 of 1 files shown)


Grygir:
Hello @sadortun,

Thank you for the improvement. The idea is great, but the solution needs to be enhanced:

1. inline JS is restricted to use
2. this code does not live in cycles of the application, it is not disposed properly
3. `ComponentManager` is not supposed to use standalone

Please consider of using  https://github.com/oroinc/platform/blob/master/src/Oro/Bundle/UIBundle/Resources/public/js/app/components/widget-component.js for rendering block with datagrid widget. It can be done with the help of `oro_widget_render` twig function, kind of

```
{{ oro_widget_render({
    'widgetType': 'block',
    'url': path('oro_datagrid_widget')
}) }}
```

The goal is to initiate block widget when it is necessary and dispose it properly when it is time.

sadortun:
@Grygir Understandable, this was only a prrof of concept :) !

sadortun:
@Grygir The goal of using an ajax request is to load the grid asynchronously, since in some case it can take up to a minute to load with very complex queries.

Using `oro_widget_render` seems to only be able to load the grid in a blocking way, therefore blocking the usage of the app until loading is completed.

Am i missing something ?

Thanks

sadortun:
Ok, after some extensive digging, i noticed that the `container` option was required to render the grid.

```html
        <div id="test-container"></div>
        <script type="text/javascript">
          require(['oroui/js/widget-manager', 'oro/block-widget', 'oro/dialog-widget'],
            function(widgetManager, BlockWidget,DialogWidget ) {

              var blockWidget = new BlockWidget({
                title: 'test',
                container: '#test-container',
                url: '{{  path('oro_datagrid_widget', {gridName: gridName, params: params, renderParams : renderParams} ) }}',
              });
              blockWidget.render();
            });
        </script>
```

This still show a full page loading spinner that prevent us from using the page, ill update this post if i find a solution.

Grygir:
No inline js-code should be used, this approach is prohibited. You can find it only in some legacy code, that we are going to refactor. Expect, the following code should do the same  
```
{{ oro_widget_render({
    'widgetType': 'block',
    'url': path('oro_datagrid_widget', {
        gridName: gridName,
        params: params,
        renderParams: renderParams
    })
}) }}
```
could you check?

It should output empty `<div/>` element with options within data attributes. Those attributes will be processed in browser and the grid will be rendered as block widget.

Grygir:
To solve issue with blocking rendering, it is better to extend `WidgetComponent::_openWidget` method

https://github.com/oroinc/platform/blob/master/src/Oro/Bundle/UIBundle/Resources/public/js/app/components/widget-component.js#L140-L190

to resolve `deferredOpen` at once, in case there's some option passed that says to do so. And cover the block with loading mask until real `widget.deferredRender` gets resolved

PR #965, opened 2019-10-21
Disable the validation of segment filter form

vtsykun:
Hi team,
All oro filters (like datetime, string and other) uses symfony forms for transforming filter data from user input input data. All default entity validator will be trigger, if filter form data contains the entity. Oro segments are also was implemented via a filter model. So when we use segments, all validators for segment entity will be trigger again and again.
https://github.com/oroinc/platform/blob/master/src/Oro/Bundle/SegmentBundle/Resources/config/validation.yml#L1-L8
It leads to duplicate sql queries and performance degradation
Selection_999(604)

Fixed https://github.com/oroinc/platform/issues/925

### DIFF
--- a/src/Oro/Bundle/SegmentBundle/Filter/SegmentFilter.php
+++ b/src/Oro/Bundle/SegmentBundle/Filter/SegmentFilter.php
@@ -118,6 +118,7 @@ class SegmentFilter extends EntityFilter
                 [],
                 [
                     'csrf_protection' => false,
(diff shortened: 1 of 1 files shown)


sadortun:
Related to https://github.com/oroinc/platform/issues/925

anyt:
Hi Uladzimir,

Thank you for the interesting idea, we'll investigate this.
Internal ticket id #BAP-19338

PR #972, opened 2019-11-14
Change misleading message

unitm-klas:
Decimal number is also a number and it does not pass this validation

### DIFF
--- a/src/Oro/Bundle/EntityExtendBundle/Resources/config/services.yml
+++ b/src/Oro/Bundle/EntityExtendBundle/Resources/config/services.yml
@@ -407,7 +407,7 @@ services:
                     -
                         - Regex:
                             pattern: '/^\-{0,1}[\d+]*$/'
-                            message: "This value should contain only numbers."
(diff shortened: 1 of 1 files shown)


mbessolov:
@unitm-klas thank you for the suggestion.
We cannot say "the whole number**s**" because only one whole number is allowed if we use "number" here meaning "a mathematical object [with certain properties]".
Another meaning of the word number - "a symbol" is commonly used in English, and is generally well understood in this context. 

> In common usage, number may refer to a symbol, a word or phrase, or the mathematical object.

https://en.wikipedia.org/wiki/Number

anyt:
@mbessolov can we change it to "integers"?

unitm-klas:
@mbessolov your reasoning is technically not correct, this validation explicitly checks for integers and they are defined as whole numbers. https://whatis.techtarget.com/definition/integer 
So it should be either a whole number or an integer as @anyt suggested. If you were pointing to the plural vs. singular than indeed singular is more precise.

PR #985, opened 2020-02-16
Use column_name attribute rather than fieldName

vytsci:
In some cases if you try to keep update structure clean and use column_name attribute inside field config doctrine schema update breaks, because this builder tries to build index based on field name rather than column name. This PR fixes this issue.


anyt:
Thank you for your contribution!
_Internal ticket id BAP-3940_

Issue #987, opened 2020-02-19
Workflow autostart issues


vytsci:
**Summary**  
We have a workflow with start_step "new" assigned, force_autostart true, we have no transition named __start__, we have another transition which has is_start true, but when we persist and flush this entity that has this specific workflow, workflow manager throws error "Step "new" of workflow "custom_flow" doesn't have allowed transition "__start__". 

**Steps to reproduce**  
Create any workflow that should autostart automatically and has no __start__ transaction.

**Actual Result**
Fatal error and workflow not started.

**Expected Result**  
Workflow should be started with step "new"

**Details about your environment**
- OroPlatform version: 3.1.19
- PHP version: 7.2
- Database: MySQL 5.7
**Additional information**  

We tried to configure __start__ transaction, even tried to remove start_step, but then nothing happens. If you leave start_step and have __start__ transaction as is_start true, then same error occurs. I always try make PR, but workflow seems so broken in so many places, for example Listener which registers which workflow to autostart, does not register anything if you have no start_step assigned, but in the docs start_step is mentioned as an optional if you have no start transition assigned. __start__ is used everywhere as start transition even if you have no such transition in your workflow. Ill try to make PR, but i need better understanding of how everything should be. For now adding __start__ to allowed transition for our start step solved an issue, but I think workflow is still broken or documented poorly.

anyt:
Hi @vytsci,
you can configure the basic workflow prototype from UI and dump it to the code with below command:
```bash
php bin/console oro:debug:workflow:definitions <workflow-name>
```

This can help you to find the initial issue in the configuration.

vytsci:
Ok we solved one issue, as database got somehow corrupted and our new entities dragged old workflow items from the table. But issues still persists where force_autostart does not work if start_step is not present. And __start__ transition should be documented inside docs.

vytsci:
I dunno this issue is partially true now. Because autostart still not works as expected, but Ive solved fatal errors and it was on our side.

PR #988, opened 2020-02-19
Workflow autostart issues #987

vytsci:
This change fixes issue where __start__ transition is defined but skipped due to missing start_step.
Also this should add all is_start transitions into consideration, but for now everywhere __start__ is hardcoded and I have no confidence in fixing multiple transitions autostarts.

### DIFF
--- a/src/Oro/Bundle/WorkflowBundle/EventListener/WorkflowStartListener.php
+++ b/src/Oro/Bundle/WorkflowBundle/EventListener/WorkflowStartListener.php
@@ -48,7 +48,10 @@ class WorkflowStartListener
         }
 
         foreach ($this->getApplicableWorkflowsForStart($entity) as $activeWorkflow) {
-            if ($activeWorkflow->getStepManager()->hasStartStep()) {
+            if (
+                $activeWorkflow->getStepManager()->hasStartStep()
+                || $activeWorkflow->getTransitionManager()->getStartTransition(null)
+            ) {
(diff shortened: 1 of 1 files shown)


vytsci:
#987 This issue is related

anyt:
👍 
Can you share the workflow configuration to reproduce the issue?

vytsci:
Sorry Im late, not currently working on this project anymore as company went down during COVID-19. 
`
imports:
    - { resource: 'claims_application_flow/steps.yml' }
    - { resource: 'claims_application_flow/attributes.yml' }
    - { resource: 'claims_application_flow/transitions.yml' }
    - { resource: 'claims_application_flow/transition_definitions.yml' }
    - { resource: 'claims_application_flow/entity_restrictions.yml' }

workflows:
    claims_application_flow:
        entity:           IfDelayed\Oro\Bundle\ClaimsBundle\Entity\Application
        entity_attribute: application
        start_step:       pending
        force_autostart:  true
`
and
`
workflows:
    claims_application_flow:
        transitions:
            __start__:
                is_start:              true
                is_hidden:             false
                is_unavailable_hidden: true
                step_to:               pending
                transition_definition: __start___definition
`

And if I remove start_step, as I want my entity to start without workflow initiation, Im unable to do so. We wanted for entity to start without workflow and initiate it via trigger when data is being properly entered, because we have lots of trash data and only small part of claims goes as legit ones, now we are forced to create workflow records for all of our claims.

PR #997, opened 2020-03-24
Notification recipients list is not as expected

vytsci:
We have encountered a bug, where doctrine metadata wont lazy load relations and in that case our notification recipients list is not as expected. Lots of users have not received their email because of this bug, it is maybe a Doctrine issue, but im adding PR just in case for someone with same bug.

When we use property accessor then we basically use getters and all the relations will be lazy loaded.


anyt:
Hi @vytsci,
Can you describe steps to reproduce the issue with not received email?

vytsci:
Yes, Oro version is 3.1.19 and Doctrine ORM 2.6.3, create workflow transition notification rule and add contact group that is association field.
image
And Doctine wont lazy load this association, this recipient will be ignored because getAssociationValue returns null. This is only a suggestion fix, but if I use property accessor seems like issue is solved as property accessor uses getters to retrieve values, and getters triggers lazy load.

PR #1007, opened 2020-04-17
Reduce logging level from 'notice' to 'info'

sadortun:
Reducing the log level from `notice` to `info` to be the same as `'Message received'` message and other informational messages produce by the Queue

### DIFF
--- a/src/Oro/Component/MessageQueue/Consumption/QueueConsumer.php
+++ b/src/Oro/Component/MessageQueue/Consumption/QueueConsumer.php
@@ -215,7 +215,7 @@ class QueueConsumer
                 'time_taken' => $executionTime
             ];
             $this->addMemoryUsageInfo($loggerContext);
-            $logger->notice('Message processed: {status}. Execution time: {time_taken} ms', $loggerContext);
+            $logger->info('Message processed: {status}. Execution time: {time_taken} ms', $loggerContext);
 
             $extension->onPostReceived($context);
         } else {


mbessolov:
@sadortun can you provide some reasoning for this request?

The commit that you referencing was in 2017, which was 3 major versions ago.
And the current log level is well aligned with the recommendations on https://github.com/Seldaek/monolog/blob/master/doc/01-usage.md#log-levels

sadortun:
Hi @mbessolov 

Thanks for your quick follow up !

A few toughs on this : 

1. I think that equally relatable logs should be on the same logging level. 

For example : 
- $logger->info('Start consuming');
- $logger->info('Message received')
- $logger->info('Message processed') 
- $logger->info('Idle');

With theses messages, i do not see why the `Message processed` message should be on a different level as `Message received` since they are both related to the same subject and to the same action ( aka processing the message)

2. As refered by the doc, `Normal but significant events.`,   and `significant`  .... `sufficiently great or important to be worthy of attention; noteworthy.`

I dont see the message `Message processed` as significant,  but more as informational or an interesting event.

3. On a logging management basis,  `Notices` are things that indicate that everything is OK, but we should still pay attention to it because of possible future issues or side-effects. For example deprecation notices.

I do not see any value to have a `Message processed` notice, since the message is to no importance, UNLESS, you are debugging or investigating issues on the message queue consumer(s).

Disclamer: This is just an opinion :) 

Have a great day,
Samuel

PR #1019, opened 2020-05-25
Point configuration provider to /Resources/config/oro

itsekhmistro-smile:
Fix location directory for `processes.yml` config files loaded in the Workflow bundle.
All `/Resources/config/*/processes.yml` are located in `/Resources/config/oro/`
Please see `find . -name processes.yml`.

This patch prevents `oro:process:configuration:load` command from failing with system error "Too many files open" due to heavy recursive directory scan.

### DIFF
--- a/src/Oro/Bundle/WorkflowBundle/Configuration/AbstractConfigurationProvider.php
+++ b/src/Oro/Bundle/WorkflowBundle/Configuration/AbstractConfigurationProvider.php
@@ -11,7 +11,7 @@ abstract class AbstractConfigurationProvider
     /**
      * @var string
      */
-    protected $configDirectory = '/Resources/config/';
+    protected $configDirectory = '/Resources/config/oro';
 
     /**
      * @var array


orocla:

It looks like this may be your first contribution to an Oro, Inc. open source project, in which case you'll need to sign a Contributor License Agreement (CLA) at https://oroinc.com/b2b-ecommerce/contributor-license-agreement/

anyt:
Hi Ivan, 
What OS do you use?

itsekhmistro-smile:
@anyt , 

Mac OS X 
```
% sysctl kern.maxfiles
kern.maxfiles: 49152
% sysctl kern.maxfilesperproc
kern.maxfilesperproc: 24576
```

:) 
For sure I know that the issue could be solved by increasing the system limits.
But system limit is not the root cause of the issue.

The config dir is not correct and the fix is quite obvious, isn't it?

anyt:
The fact that in a core we have processes.yml files only in an `oro` folder doesn't mean someone else in customization hasn't added these files outside of it. 
It could be a huge BC break, so I can't apply it to the 4.1 branch. That's why I asked you about an OS to see is there a workaround. And as you said, on Mac you can increase the max files limit, and as I know, on Linux it's not an issue at all.

For the master branch we can fix it for sure.

itsekhmistro-smile:
>For the master branch we can fix it for sure.

Thank you. I'm the same with you about versioning.

Just wanna make the world a better place.

PS. On Linux it could be easily reproduced as well:
```bash -c "ulimit -n 1000 &&  bin/console ...```
It's just that the default values for those limits are higher on Linux than on Mac.
I have a non-default value (increased one) on my Mac.

PR #1023, opened 2020-06-16
Add translatable email templates fixture

klas:
Add a missing functionality to load translated email templates. 

Localizations are derived from the file names following localization .ini files convention
**template_name.locale_code.format**
e.g. 
customer_user_confirmation_email.en.html
customer_user_confirmation_email.de_DE.html


mbessolov:
@klas, thank you for the pull request.

Considering that this abstract fixture is not used anywhere in OroPlatform, it should be covered with automated tests, or else we won't be able to track if it works or breaks in the future versions.

This fixture seems to silently ignore templates for non-active localizations.

There also seems to be a problem with filenames that have more than 2 dots (e.g. some.custom.template.de_DE.html).

PR #1044, opened 2020-10-17
Add user to oro_user_organization table

sadortun:


anyt:
The fix is merged. Thank you for your contribution!

PR #1048, opened 2020-11-03
BUGFIX: issue with dump translation due to method name change into \S…

sylvainraye:
…ymfony\Component\Translation\Writer\TranslationWriter


anyt:
I'm closing the PR, because the following functionality was removed in 4.2 LTS, as we don't want to keep it the part of the application. 
See https://github.com/oroinc/platform/blob/1a1ed7f9faff484d32538ef06c090813804f0d21/incompatibilities-4-2.md#translationbundle.

sylvainraye:
oh, this looks like a bad news for the Oro Deepl Integration I did !!!

mbessolov:
@sylvainraye I do not see any references to the removed services in https://github.com/diglin/oro-deepl, so it does not look like your integration will be affected.

sylvainraye:
oh, thank you, I wanted to check that later. You were faster. Thank you very much. I was afraid that the 4.2 may have an influence on this bundle. I will still test it to be sure but I'm glad to know that at first sight, no issue

PR #1054, opened 2021-02-24
EmailTemplateCompilationException wrong description message

yaroslavbr:
- fixed exception message

### DIFF
--- a/src/Oro/Bundle/EmailBundle/Exception/EmailTemplateCompilationException.php
+++ b/src/Oro/Bundle/EmailBundle/Exception/EmailTemplateCompilationException.php
@@ -14,7 +14,10 @@ class EmailTemplateCompilationException extends EmailTemplateException
      */
     public function __construct(EmailTemplateCriteria $criteria)
     {
-        $message = sprintf('Could not found one email template with "%s" name', $criteria->getName());
+        $message = sprintf(
+            'Compilation error during rendering template "%s". Please look at the log file for more details.',
+            $criteria->getName()
+        );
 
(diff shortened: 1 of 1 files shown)


src/Oro/Bundle/EmailBundle/Exception/EmailTemplateCompilationException.php
```diff
@@ -14,7 +14,10 @@ class EmailTemplateCompilationException extends EmailTemplateException
      */
     public function __construct(EmailTemplateCriteria $criteria)
     {
-        $message = sprintf('Could not found one email template with "%s" name', $criteria->getName());
+        $message = sprintf(
+            'Compilation error during rendering template "%s". Please look at the log file for more details.',
```
anyt, line 18: > Please look at the log file for more details.

As on a production web server, this message will be already displayed in a log. It's odd to see this recommendation in the exception message. Also, not all applications use files to store logs. And we never recommend the same for any other exception messages, so I propose to remove the second part. 

Instead, we can provide more details about an error in the exception itself. For example, now we don't pass the original exception to the parent Exception constructor. While in \Oro\Bundle\EmailBundle\Provider\EmailTemplateContentProvider::getTemplateContent, we do have the original exception, and we can add it to the stack trace, so the developer will not need to check the log file in the `dev` environment, and the system administrator will see the message in the original trace, that is preferred, instead of having two unrelated log messages, that could even not go in a row, in case of a high load.

anyt (changes requested):

Issue #1061, opened 2021-05-11
Oro ignore-tabbable.js breaks default tab to focus behaviour


jimohalloran:
**Summary**  
The default behaviour for a HTML based web app is to highlight the contents of a field when tabbing into it so the value can be replaced easily.  This is what I (and my users) expect Oro to do as a standard web application.  However, since 4.1.11 (or 12, hard to tell), when I tab between fields the cursor moves as expected, but any existing value is not highlighted.  I'd like the default highlight behaviour to be restored.

**Steps to reproduce**  
Open a form with multiple text fields.  Enter values into each.  Click on the first text field to give it focus.  Then hit tab to move to the next.  Note how the cursor moves but the text content is not selected.

**Actual Result**
When I tab between fields, the cursor moves as expected, but the value is night selected.

**Expected Result**  
When I tab between fields, the cursor should move to the next field, and any value in that field should be selected to allow quick/easy over typing to replace the value.  You can see the default behaviour here: https://getbootstrap.com/docs/4.0/components/forms/#sizing and note how the existing value is highlighted when you tab between fields.

**Details about your environment**
- OroPlatform version: 4.1.12
- PHP version: N/A Caused by JS
- [Optional] Client operating system Linux - Centos Stream 8
- [Optional] Client browser and its version: Chrome 90

**Additional information**  
I can see the default behaviour (existing value is highlighted on focus) on a dead simple HTML form without any JS that I create in a file and open in the browser.  I can also see it in the sample forms in the Bootstrap docs.  So I'm confident that the latest Chrome releases haven't changed the default.  I can repro on both my own custom forms, and standard forms like customers.  I can also repro in the OroCRM online demo, so it's not related to anything I've customised.

Digging into the issue further it seems to be caused by an OroUI component (https://github.com/oroinc/platform/blame/master/src/Oro/Bundle/UIBundle/Resources/public/js/app/modules/ignore-tabbable.js) which is manually managing focus in response to tab key presses.  If I use devtools to remove this listener from a form element, the normal focus behaviour works as expected.  

I'm guessing the lack of highlight is due to this code setting focus on the new control then preventing the default behaviour without manually highlighting the content?  

It does seem like this would have been an Oro 4.1.11 or 4.1.12 change though from the dates, although I can't really find anything in the release notes to confirm that.

Which I guess leaves two questions.  Why does this thing exist?  And how do I either disable it, or fix it so I still get the default browser behaviour I expect?

jimohalloran:
[...]

PR #1062, opened 2021-05-17
Oro ignore-tabbable.js breaks default tab to focus behaviour

jimohalloran:
This PR is a simple fix for #1061.  Currently when ignore-tabbable.js switches focus to a new element, the contents of the new element is not highlighted breaking the default web browser behaviour.

This PR checked to see if the element just focused has a "select" method, and if so calls it to select the contents of the newly focused element.  Select boxes and buttons don't have a selection and therefore we shouldn't just blindly call `select()` on everything.


anyt:
Thank you for your contribution, @jimohalloran.
The team will review your PR in the scope of the above bug fix and merge if it suits.
_Internal ticket id BB-20605_.

PR #1100, opened 2023-05-15
Fix SegmentBundle migration to support MySQL platform as well as PgSQL

idimsh:
Fix migration v1.11 in SegmentBundle to support MySQL platform as well as the current PgSQL with its plain query execution.

* Make the migration connection aware
* Tested so that migrating from 4.2 on MySQL to 5.1 (on MySQL as well) will not failed due to this migration plain query, this fix will allow the execution of queries against the current MySQL before any planned PgSQL platform change on running systems.


anyt:
Hi @idimsh, 
Unfortunately, MySQL is not supported anymore for OroPlatofrm. You have to [migrate the database to PostgreSQL](https://doc.oroinc.com/backend/setup/mysql-to-postgresql) before the upgrade.

PR #1105, opened 2023-09-25
Create console command to view global config

carbsrule:
Currently an administrator can update the Oro config via the CLI using `oro:config:update`

This adds `oro:config:view` which enables viewing the current config

### DIFF
--- a/src/Oro/Bundle/ConfigBundle/Command/ConfigViewCommand.php
+++ b/src/Oro/Bundle/ConfigBundle/Command/ConfigViewCommand.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Oro\Bundle\ConfigBundle\Command;
+
+use Oro\Bundle\ConfigBundle\Config\ConfigManager;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Views a configuration value in the global scope.
+ */
+class ConfigViewCommand extends Command
+{
+    protected static $defaultName = 'oro:config:view';
+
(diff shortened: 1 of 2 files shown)


mbessolov:
Hi @carbsrule

Thank you for your contribution, it could be quite useful for some developers/adminstrators.

In order for this to be accepted though, it needs some additional changes:

1. Properly support output of different data types. If some value cannot be pretty-printed, the command should indicate it.
2. It should properly handle (i.e. not show) sensitive values - if the value is masked in the system configuration UI, it should be masked in this command's output as well.
3. Tests - please look at how other commands are tested. Most likely we will need a functional test and a unit test to properly cover the (updated version of) this command.

Thank you

carbsrule:
@mbessolov 

1. Pretty printing is simple enough and I have (uncommitted) code to handle array|object|bool configs
2. Seems possible using `oro_config.provider.system_configuration.form_provider` (`Oro\Bundle\ConfigBundle\Provider\SystemConfigurationFormProvider`) to work out how the field is configured, although TBH I'm not sure if this is desired behaviour. The `oro:config:update` command is completely unaware of the system config UI, and thus, for example, will dump a plain text value into a field which is meant to be encrypted, so for me it makes sense for this mirror command to behave in a similarly unaware fashion. A user who has access to the CLI likely has access to the DB anyway, so this is just (meant to be) a convenient way of grabbing the value without needing to write an SQL query or manually decode the value. Thoughts?
3. I haven't looked at this yet but can do so

carbsrule:
@mbessolov I've blocked access to encrypted fields (a possible improvement would be to support a `--decrypt` flag which would display an encrypted value as though the data were unencrypted), and have added unit and functional tests. I think this is ready, at least for my purposes.

mbessolov:
@carbsrule thank you, looks great.

Please review and sign the contributor license agreement at https://oroinc.com/b2b-ecommerce/contributor-license-agreement/ so that we can merge you contribution to Oro codebase.

PR #1140, opened 2026-04-17
Fix nullable return in getUserIdentifier() when username is not set

Dylvn:
Fixes https://github.com/oroinc/customer-portal/issues/14

## Problem

When importing `CustomerUser` records without an ID (creation) and without an email address, a `TypeError` is triggered during Symfony validation.

The validator calls `__toString()` on the freshly instantiated entity, which internally calls `getUserIdentifier()`:

    public function __toString()
    {
        return (string)$this->getUserIdentifier();
    }

The `(string)` cast never executes. The `TypeError` is raised inside `getUserIdentifier()` itself, before the cast can apply.

## Root cause

`$username` is declared as `?string` at the Doctrine level, which is a valid state during entity construction (e.g. import processing before the field is hydrated). However, Symfony's `UserInterface` contract enforces a strict `string` return type on `getUserIdentifier()`.

## Fix

Return an empty string as fallback to bridge the gap between the nullable Doctrine mapping and the strict Symfony interface contract.

### DIFF
--- a/src/Oro/Bundle/UserBundle/Entity/AbstractUser.php
+++ b/src/Oro/Bundle/UserBundle/Entity/AbstractUser.php
@@ -142,7 +142,7 @@ abstract class AbstractUser implements
     #[\Override]
     public function getUserIdentifier(): string
     {
-        return $this->username;
+        return $this->username ?? '';
     }
 
     public function setUserIdentifier($username): self


mbessolov:

It looks like this may be your first contribution to an Oro, Inc. open source project, in which case you'll need to sign a Contributor License Agreement (CLA) at https://oroinc.com/b2b-ecommerce/contributor-license-agreement/

If you've already signed the CLA, it's possible we don't have your GitHub username or you're using a different email address. GitHub uses the email address you set in your local Git configuration to associate commits with your GitHub account. Please sign the CLA again using the correct GitHub username and email address or see this help article on [setting the email on your git commits](https://docs.github.com/en/account-and-profile/how-tos/email-preferences/setting-your-commit-email-address).

PR #1143, opened 2026-04-30
Allow pnpm to install packages

jimohalloran:
In the transition from npm to pnpm the changeover from running `npm ci` to `pnpm i --lockfile-only` resulted in incorrect behaviour.

The `npm ci` command downloaded and installed packages at the specific versions as specified in the lockfile.

`pnpm i --lockfile-only` on the other hand does not download dependencies, it just updates the lockfile.  This is not a substitute for `npm ci`.

What we want instead of `pnpm i --frozen-lockfile` as this will download and install exactly the dependencies specified in the lock file, which is equivelent to the `npm ci` command.

The problem with using `--lockfile-only` is that when trying to deploy outside of dev or CI environments (`--frozen-lockfile` is by default on in CI) no npm dependencies are downloaded or installed, which results in a broken system.

This patch updates the command line used by the composer helper script accordingly.

### DIFF
--- a/src/Oro/Bundle/InstallerBundle/Composer/ScriptHandler.php
+++ b/src/Oro/Bundle/InstallerBundle/Composer/ScriptHandler.php
@@ -349,7 +349,7 @@ class ScriptHandler
     private static function pnpmCi(IOInterface $inputOutput, int $timeout = 60, bool $verbose = false): void
     {
         $logLevel = $verbose ? 'info' : 'error';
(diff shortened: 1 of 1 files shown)


mbessolov:
@jimohalloran Thank you for your contribution!
The proposed changes will be merged and included in an upcoming release.

mbessolov:
_internal ticket - BB-27327_

